PR #20, opened 2020-06-29
Add Dash into AIPscan

ross-spencer:
Initial attempt to integrate Dash.


ross-spencer:
Closing, but not deleting the branch as we don't need this piece just yet (unlikely at all, but this work also isn't in source control elsewhere).

Issue #35, opened 2020-07-10
Introduce test runners (tox), and travis configuration for CI


ross-spencer:
Just some basic repository health imported from other Artefactual repositories.

ross-spencer:
* Part 1: https://github.com/artefactual-labs/AIPscan/pull/36 

I'll prepare a Travis commit but I think that should be added later as it will generate too much noise for the refactor at the moment, i.e. returning Flake8 bits n pieces we already know about.

ross-spencer:
* Part 2: https://github.com/artefactual-labs/AIPscan/issues/35

ross-spencer:
Because circular imports, we haven't quite got here yet! But close:
```
AIPscan/__init__.py:10:1: E402 module level import not at top of file
AIPscan/__init__.py:10:1: F401 'celery.Celery' imported but unused
AIPscan/__init__.py:11:1: E402 module level import not at top of file
AIPscan/__init__.py:20:1: E402 module level import not at top of file
AIPscan/__init__.py:20:1: F401 'AIPscan.models' imported but unused
AIPscan/__init__.py:22:1: E402 module level import not at top of file
AIPscan/__init__.py:23:1: E402 module level import not at top of file
AIPscan/__init__.py:24:1: E402 module level import not at top of file
AIPscan/__init__.py:25:1: E402 module level import not at top of file
```

ross-spencer:
cc. @peterVG @tw4l I spent some time trying a few ways to get `AIPscan/__init__.py` to pass - and didn't quite find the right approach, so a second pair of eyes might help. 

If that turns out to be too onerous, then a temporary approach might be to make a `flake8` expception for this file, merge https://github.com/artefactual-labs/AIPscan/issues/35 and then we'll have CI available within GitHub for the rest of the project which will be beneficial for objective review, and generally keeping quality up.

ross-spencer:
Closed by https://github.com/artefactual-labs/AIPscan/pull/101/commits

PR #36, opened 2020-07-10
Add tox automation

ross-spencer:
Running tox will enable tests to be picked up and run automatically
via Pytest and run Flake8 conformance and Black linting initially.

**For the reviewer**

The exceptions in the tox config are largely about making `black` happy and so we can use it to its full potential. We adopt similar exclusions in [amclient](https://github.com/artefactual-labs/amclient/blob/25bafd9f8c2831f06663f966216779c9f7fe4851/tox.ini#L18-L23). 

Pytest will currently exit with exit code 5. This is simply because there are no tests yet. 

Connected to https://github.com/artefactual-labs/AIPscan/issues/35


peterVG:
Good to add this. Thanks @ross-spencer

Issue #48, opened 2020-08-13
Provide a report/view that totals all package types per storage service.


peterVG:
Total number of packages in Storage Service at a given date, with a breakdown of:
-AIPs
-replicated AIPs
-deleted AIPs
-DIPs
-transfers

ross-spencer:
We need to add a summary of this. A chart above the table here might be a sensible place? 

![image](https://user-images.githubusercontent.com/1880412/92960751-28538a00-f43c-11ea-8c01-266cdc39e720.png)

In the meantime the schema has been updated here with the counts output to the database as well: https://github.com/artefactual-labs/AIPscan/pull/51/commits/69278cbedf8e828148f76f4c0906bffb77b51776

Issue #47, opened 2020-08-13
Provide a breakdown of all packages in SS, not just the AIPs included in reports


peterVG:
Include a log of the Exception errors that were reported for each.

Provide this count on the reports themselves?

joel-simpson:
This could also be framed as a report on what AIPScan includes in its database and what it doesn't. That's not just about errors (i.e. couldn't scan x AIPs) but also about helping users reconcile what is in AIPScan with what is in their storage system, including all packages, including DIPs, replicated AIPs, transfers, SIPs etc. 
One idea we discussed was providing an explicit way for users to reconcile what they see in the Storage Service (whether through SS UI or the Archival Storage tab) with what is included in AIPScan. This should give users confidence in the accuracy of data stored in AIPScan.

ross-spencer:
I have got as far as cleaning up some of the METS work to make it easier to see where we need to log these issues. What I haven't  figured out yet, is where to record them:

1. An errors table with a foreign key to the AIP and files (original/preservation) causing a problem?
2. Errors in the AIP or 'files' tables?

And we need to add a summary to the storage service view as well. 

I believe there are currently two places to log errors in processing:

1. Overall error processing the METS: [here](https://github.com/artefactual-labs/AIPscan/blob/852e3d25ab3a78e27870e5b1c798a5f91953ffb8/AIPscan/Aggregator/tasks.py#L286-L290)
2. Problem processing a file: [here](https://github.com/artefactual-labs/AIPscan/blob/852e3d25ab3a78e27870e5b1c798a5f91953ffb8/AIPscan/Aggregator/database_helpers.py#L184-L188). The problem with this particular except right now is that the error assumes ISO file formats but it could really be any format, and so it might be better to log the entire file as something like "not able to process" and then move on.

peterVG:
To spit up Joel's recommendation to have a way to report out which SS packages have been left out of AIPscan (e.g. DIPs, deleted packages, replicated packages, transfers in backlog) from any error handling functionality, I've created a new issue for the former. The way it's worded (and the recommended implementation) is inspired by the thinking around Issue https://github.com/artefactual-labs/AIPscan/issues/95.

See Issue https://github.com/artefactual-labs/AIPscan/issues/110

peterVG:
GUI option 1: display stats in small, greyed out text below total package count.

[...]

PR #51, opened 2020-09-04
Packages + METS loading refactor, tests, and logging

ross-spencer:
Connected to #47 
Connected to #48


ross-spencer:
@peterVG there's no particular rush on a review for this. But it might be of interest to you to read review. 

The work represents a decent point to review and move onto other bits and pieces within AIPscan. The code largely represents stabilizing the Aggregator module a little more, adding tests, and getting a bit more information into the database around packages. Lots more to think about, I'll head back in Monday and will be thinking about stepping up work in the data module and writing the views for those. We'll probably all touch base early on next week I'm guessing.

Issue #52, opened 2020-09-17
Derivatives-originals report is broken in select cases


ross-spencer:
When trying to access derivative information some storage services are returning an AttributeError where `.file_fomat` is `None`.

ross-spencer:
The first cause of this issue, I believe may be down to certain transfer types having different derivation types, so Dataverse has a `fileGrp USE` of `derivative` which is created from their tabular imports. Preservation copies have `preservation` which I think we already deal with, and access copies have no visible output in the AIP METS so I don't know if there are more to capture, but let's start with this and see... 
```xml
    <mets:fileGrp USE="derivative">
      <mets:file GROUPID="Group-20439035-0516-4bbc-9ab9-fd4a4c56033f" ID="file-eca9811b-e240-4372-b7a7-f4eda44c0d8f" ADMID="amdSec_2">
        <mets:FLocat xlink:href="objects/afternoon-snacks-1.RData" LOCTYPE="OTHER" OTHERLOCTYPE="SYSTEM"/>
      </mets:file>
      <mets:file GROUPID="Group-20439035-0516-4bbc-9ab9-fd4a4c56033f" ID="file-953418f2-fce3-4e78-817c-c8123c835351" ADMID="amdSec_4">
        <mets:FLocat xlink:href="objects/afternoon-snacks-1.tab" LOCTYPE="OTHER" OTHERLOCTYPE="SYSTEM"/>
      </mets:file>
    </mets:fileGrp>
  </mets:fileSec>
```

Issue #55, opened 2020-09-17
Originals and Copies database tables could be collapsed into single Files table


tw4l:
As can be seen in the ERD in https://github.com/artefactual-labs/AIPscan/issues/42, the AIPscan database currently has separate tables for original files (`originals`) and preservation derivatives (`copies`). The majority of fields for the two tables are identical, and it seems to me that they could be collapsed into a single `files` table.

This would make the development experience a bit smoother. As an example, to get information about the largest files, we currently have to query two tables:

```
if original_files is True:
    largest_files = originals.query.order_by(originals.size.desc()).limit(20)
else:
    largest_files = copies.query.order_by(copies.size.desc()).limit(20)
```

I'd prefer to be able to do something like:

```
largest_files = files.query.filter_by(type=file_type).order_by(files.size.desc()).limit(20)
```

Or, to get the largest files regardless of type, simply:

```
largest_files = files.query.order_by(files.size.desc()).limit(20)
```

We could manage the file type via a `type` field controlled by an enumerated list, or with a `is_original` boolean (the enumerated list might age better if we anticipate eventually tracking information about other types of files such as metadata and access derivatives in addition to original files and preservation derivatives).

ross-spencer:
It's a plus one from me for simplicity. I do think this can be controlled by an enumerated list and that it might help in other reports too. You can see [here](https://github.com/artefactual-labs/AIPscan/blob/da8fc4c939a927cf0dbd54d7ab2fedc6f625e140/AIPscan/Data/data.py#L66-L69) I want to treat both types very similarly, but there is a small difference with there being a PUID or not. 

Do you see any problems creating a relationship to multiple derivatives though Tessa - in PR56 just logged, the dataverse original has two derivatives, for arguments sake, say a CSV and TAB. Does that make sense in a flatter structure, or should other parts of the files table sit independently from the rest? (one to many to a relationship lookup table?) (I can't quite see it in my head yet, but it's the end of the day, and a clue to head home for me!)

tw4l:
I think multiple derivatives would still work well in a flatter table structure. As an overly simplified illustration (assuming a custom preservation rule that converts xslx to ods for preservation, just for demonstration), say we have a few types of derivatives for a single original file. The table could look something like:

| id | name | type | format | puid | derivative_of |
| --- | --- | --- | --- | --- | --- |
| 1 | my-research-data.xlsx | original | Microsoft Excel for Windows | fmt/214 | null |
[...]

PR #56, opened 2020-09-17
WIP: Add fileGrp USE="Derivative" handling to fetch

ross-spencer:
Fetch jobs were not successfully picking up the derivative type which
is used by Dataverse transfers. We add that here.

Connected to artefactual-labs/aipscan#52

### DIFF
--- a/AIPscan/Aggregator/database_helpers.py
+++ b/AIPscan/Aggregator/database_helpers.py
@@ -142,12 +142,32 @@ def process_aip_data(aip, aip_uuid, mets):
     need for reporting.
     """
 
+    # we should only see Directory and "File" types in this field.
+    DIRECTORY_TYPE = "Directory"
+
+    # Objects grouped by use in the fileGrp.
     ORIGINAL_OBJECT = "original"
     PRESERVATION_OBJECT = "preservation"
+    METADATA_OBJECT = "metadata"
+    SUBMISSION_OBJECT = "submissionDocumentation"
+
+    # Derivative is a USE introduced in the Dataverse analysis.
+    DERIVATION_OBJECT = "derivative"
+
+    # Tuples of objects to make comparison easier below.
+    KNOWN_UNUSED_TYPES = (METADATA_OBJECT, SUBMISSION_OBJECT)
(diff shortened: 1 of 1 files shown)


ross-spencer:
@peterVG @tw4l what do you think about this fileGrp handling in AIPscan. The derivative type expresses a relationship between a tabular format in Dataverse and the concept of a Bundle which is described in the [Wiki](https://wiki.archivematica.org/Dataverse#Bundles_for_tabular_data_files). (Tessa, you probably have more institutional knowledge of this!) 

We noticed the original-derivatives report wasn't being written out correctly. While I happened to use derivative as the term in that report, the other parts of AIPscan are focused on either originals or preservation copies:

image

image

So, that leads me to ask, what is the full extent of the change I might make for AIPscan to handle derivatives properly. Largely, they are in the camp of:

1. Treat them as unique, and update the screens (and maybe add a derivative type to the report). 
2. Treat them as originals, and leave the screens the same, and don't talk about them separately. 
3. ~~Treat them as preservation copies~~ (As ever putting all options out there, this one doesn't feel correct).
4. Something else?

ross-spencer:
Ah! Just going to read this Tessa! https://github.com/artefactual-labs/AIPscan/issues/55 (which seems timely!)

tw4l:
@ross-spencer It's an interesting question!

Are these Dataverse derivatives unique in the respect of being derivatives that are not necessarily preservation derivatives and were created in advance of ingest into Archivematica? It seems that the choice to use fileGrp="derivative" signifies an intent not to think of them in Archivematica as original files or preservation copes (I agree with you that the latter seems particularly not right).

I'm wavering between 1 and 2, and might suggest a slightly hybrid approach:

- Store them in the database as a distinct type of file (maybe type="dataverse_derivative" or type="derivative", to tie it to my suggestions in #55) to retain granularity of the data.
-Decide whether to treat them as originals or as distinct on a per-report basis. This might be where some user stories would really help us out. If as an administrator looking for reports my intention is to separate out the files Archivematica received from the files Archivematica created, I could see a use case for considering the Dataverse derivatives as an original file in some cases. In some other cases, it might make sense to view them separately.

tw4l:
Closing due to inactivity and because this branch has drifted from main.

Issue #57, opened 2020-09-18
AIP original name isn't always in the first dmdSec


ross-spencer:
We need a more sophisticated search mechanism.

Related to https://github.com/archivematica/Issues/issues/1305

ross-spencer:
An example of the complexity here by documenting empty directories:
```xml
  <mets:dmdSec ID="dmdSec_1">
    <mets:mdWrap MDTYPE="PREMIS:OBJECT">
      <mets:xmlData>
        <premis:object xmlns:premis="http://www.loc.gov/premis/v3" xsi:type="premis:intellectualEntity" xsi:schemaLocation="http://www.loc.gov/premis/v3 http://www.loc.gov/standards/premis/v3/premis.xsd" version="3.0">
          <premis:objectIdentifier>
            <premis:objectIdentifierType>UUID</premis:objectIdentifierType>
            <premis:objectIdentifierValue>e944bddb-0cc0-4005-ae30-4a5b3ee84d9f</premis:objectIdentifierValue>
          </premis:objectIdentifier>
          <premis:originalName>%transferDirectory%objects/E1/</premis:originalName>
        </premis:object>
      </mets:xmlData>
    </mets:mdWrap>
  </mets:dmdSec>
  <mets:dmdSec ID="dmdSec_2">
    <mets:mdWrap MDTYPE="PREMIS:OBJECT">
      <mets:xmlData>
        <premis:object xmlns:premis="http://www.loc.gov/premis/v3" xsi:type="premis:intellectualEntity" xsi:schemaLocation="http://www.loc.gov/premis/v3 http://www.loc.gov/standards/premis/v3/premis.xsd" version="3.0">
          <premis:objectIdentifier>
            <premis:objectIdentifierType>UUID</premis:objectIdentifierType>
            <premis:objectIdentifierValue>4a3773b4-1d87-4a4a-aedc-1c1f8fa585fa</premis:objectIdentifierValue>
          </premis:objectIdentifier>
          <premis:originalName>%transferDirectory%objects/0B/E1/</premis:originalName>
        </premis:object>
      </mets:xmlData>
    </mets:mdWrap>
  </mets:dmdSec>
  <mets:dmdSec ID="dmdSec_3">
    <mets:mdWrap MDTYPE="PREMIS:OBJECT">
      <mets:xmlData>
[...]

Issue #53, opened 2020-09-17
Epic: Upstream issues with mets-reader-writer (metsrw)


ross-spencer:
_Please edit the top-level description here to add more issues while this ticket isn't managed by Codetree_

**Bugs**

* https://github.com/archivematica/Issues/issues/1277
* https://github.com/archivematica/Issues/issues/1129

**Enhancements**

* https://github.com/archivematica/Issues/issues/1305

PR #59, opened 2020-09-22
Improve retrieval of original name

ross-spencer:
Having used AIPscan a little more, leaving the name blank is strange
to look at, but we also cannot always retrieve the original name. We
do our best to a) retrieve the name in this commit, as well as b)
populate the name with the transfer UUID as an alternative.

Connected to artefactual-labs/aipscan#57

### DIFF
--- a/AIPscan/Aggregator/mets_parse_helpers.py
+++ b/AIPscan/Aggregator/mets_parse_helpers.py
@@ -44,36 +44,43 @@ def parse_mets_with_metsrw(mets_file):
 
 
 def get_aip_original_name(mets):
-    """Retrieve PREMIS original name from a METSDocument object."""
+    """Retrieve PREMIS original name from a METSDocument object.
+
+    If the original name cannot be reliably retrieved from the METS file
+    a METSError exception is returned to be handled by the caller as
+    desired.
+    """
 
     # Negated as we're going to want to remove this length of values.
     NAMESUFFIX = -len("-00000000-0000-0000-0000-000000000000")
 
+    # The transfer directory prefix is a directory prefix that can also
+    # exist in a dmdSec intellectual entity and we want to identify and
+    # ignore those.
+    TRANSFER_DIR_PREFIX = "%transferDirectory%"
+
     NAMESPACES = {u"premis": u"http://www.loc.gov/premis/v3"}
     ELEM_ORIGINAL_NAME_PATTERN = ".//premis:originalName"
 
-    FIRST_DMDSEC = "dmdSec_1"
-
     original_name = ""
     for fsentry in mets.all_files():
-        try:
-            dmdsec = fsentry.dmdsecs[0]
-            if dmdsec.id_string != FIRST_DMDSEC:
-                continue
+        for dmdsec in fsentry.dmdsecs:
             dmd_element = dmdsec.serialize()
             full_name = dmd_element.find(
                 ELEM_ORIGINAL_NAME_PATTERN, namespaces=NAMESPACES
             )
+            if full_name is not None and full_name.text.startswith(TRANSFER_DIR_PREFIX):
+                # We don't want this value, it will usually represent an
+                # directory entity.
+                continue
             try:
                 original_name = full_name.text[:NAMESUFFIX]
             except AttributeError:
-                pass
-            break
-        except IndexError:
-            pass
+                continue
 
+    # There should be a transfer name in every METS.
     if original_name == "":
-        raise METSError()
+        raise METSError("Cannot locate transfer name in METS")
 
     return original_name
 
--- a/AIPscan/Aggregator/tests/test_mets.py
+++ b/AIPscan/Aggregator/tests/test_mets.py
@@ -13,19 +13,32 @@ FIXTURES_DIR = "fixtures"
 
 
 @pytest.mark.parametrize(
-    "fixture_path, transfer_name",
+    "fixture_path, transfer_name, mets_error",
     [
-        (os.path.join("features_mets", "features-mets.xml"), "myTransfer"),
-        (os.path.join("iso_mets", "iso_mets.xml"), "iso"),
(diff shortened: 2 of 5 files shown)


AIPscan/Aggregator/mets_parse_helpers.py
```diff
@@ -49,31 +49,32 @@ def get_aip_original_name(mets):
     # Negated as we're going to want to remove this length of values.
     NAMESUFFIX = -len("-00000000-0000-0000-0000-000000000000")
 
+    # Other intellectual entities exist in the METS, i.e. for empty
+    # directories. We can identify those with an invalid name prefix.
```
tw4l, line 53: It might be worth adding a comment that this name prefix is an indicator of an invalid `<premis:originalName>` value specifically within the context of a dmdSec. I was confused for a minute because the `<premis:originalName>` value in the `amdSec` of an original file also begins with "%transferDirectory%".

AIPscan/Aggregator/tests/test_mets.py
```diff
@@ -13,19 +13,30 @@
...
 
 @pytest.mark.parametrize(
-    "fixture_path, transfer_name",
+    "fixture_path, transfer_name, mets_error",
     [
-        (os.path.join("features_mets", "features-mets.xml"), "myTransfer"),
-        (os.path.join("iso_mets", "iso_mets.xml"), "iso"),
+        (os.path.join("features_mets", "features-mets.xml"), "myTransfer", False),
+        (os.path.join("iso_mets", "iso_mets.xml"), "iso", False),
+        (os.path.join("original_name_mets", "dataverse_example.xml"), "", True),
```
tw4l, line 20: Am I understanding correctly that this test raises a `METSError` because Archivematica has written two `dmdSec`s with an ID of `"dmdSec_1"` and mets-reader-writer isn't able to read two `dmdSec`s with the same ID? Forgive me if this naive, but with the naked eye it looks like like the `<premis:originalName>` value in the second `<mets:dmdSec ID="dmdSec_1">` (starting on line 39) should be what we're looking for.

tw4l (changes requested):
Hi @ross-spencer, good investigative work here! I think you've hit on a good solution and the code looks good to me. I've left one suggestion for a clarifying comment and a question about one of the tests.

I see that you've opened a related [mets-reader-writer](https://github.com/archivematica/Issues/issues/1305) issue. Forgive me if this is a naive question, but would it be better to make these changes to `mets-reader-writer` (i.e. add a helper to retrieve the transfer's original name, and more gracefully handle multiple `dmdSec`s with identical IDs) and then call the new helper here in AIPscan?

AIPscan/Aggregator/tests/test_mets.py
ross-spencer: That is correct. I will add notes to the test. Related issue: https://github.com/archivematica/Issues/issues/1310

ross-spencer:
Thanks for the review @tw4l. You are somewhat right about the upstream issues. We had already determined an approach to handling those which was to not look at `mets-reader-writer` in the sprint and deliver value via the application itself. I don't feel the reactions to `mets-reader-writer` in this code are necessarily mutually exclusive of fixes there. 

I have started an EPIC for us to build on in this repository: https://github.com/artefactual-labs/AIPscan/issues/53 and creating related issues in archivematica/issues. I feel we're approaching critical mass with regards to outstanding `mets-reader-writer` work so, as per the last review, suspect we'll be tackling those soon. 

I have added comments to the above and hopefully made the intentions in the code and test clearer,,. but I couldn't see what explicit changes you wanted? 

Just a note on marking "request for changes" in a PR. I recognize not everyone approaches our work like this - but we do have branch protection on this project so it's already explicit the PR can't be merged without approval - it try not to tell folks that's doubly the case. My personal approach is to leave a PR conversational by submitting comments unless there is something critically broken or out of keeping with our contributing gude that we should avoid at all costs. I tend to comment without explicit approval to create better understanding, ask questions, and that in most cases does result in some change to code. If in the rare case someone following their process approves a PR in the middle of a conversation and I am the first reviewer then I am comfortable with that as long as I know I've caught the things I had the capacity to catch that we can't really live with.

AIPscan/Aggregator/mets_parse_helpers.py
```diff
@@ -44,14 +44,20 @@ def parse_mets_with_metsrw(mets_file):
 
 
 def get_aip_original_name(mets):
-    """Retrieve PREMIS original name from a METSDocument object."""
+    """Retrieve PREMIS original name from a METSDocument object.
+
+    If the original name cannot be reliably retrieved from the METS file
+    a METSError is exception is returned to be handled by the called as
```
tw4l, line 50: I think there's a typo in "METSError is exception is returned"

tw4l (approved):
Thanks @ross-spencer. I'm on board delivering value in the application itself and keeping track of potential upstream work for now (thanks for starting that Epic - that'll prove very helpful). I think that I had not yet been in a sprint meeting when I did the last review, so I was missing a bit of the context there.

You're right that I wasn't explict about any changes. I was trying to suggest (perhaps too implicitly) that upstream changes to `mets-reader-writer` to more gracefully handle the case of duplicative dmdSec IDs might result in downstream changes here - for instance, in the `dataverse_example` test returning the original value from the second `"dmdSec=1"` rather than throwing a `METSError`.

But given the sprint priorities and current focus, I'm quite happy to remove myself as an obstacle for merging - this is good work and definitely adding value. Thanks for adding the comments - I left one additional comment on a docstring typo but otherwise looks great.

Thanks for the PR process note as well - we do seem to be a bit inconsistent as a company in how/whether we use "request for changes" but I see your point and agree that a comment without explicit approval one way or the other would have been appropriate here.

AIPscan/Aggregator/mets_parse_helpers.py
ross-spencer: Ah, good spot. Thanks Tessa.

Issue #54, opened 2020-09-17
Add report on largest files in Storage Service


tw4l:
Report description from requirements: "I want to see which is the largest file (in GB / TB) in my e-depot system"

tw4l:
WIP PR for demonstration: https://github.com/artefactual-labs/AIPscan/pull/61

PR #61, opened 2020-09-22
Add largest files report

tw4l:
Connected to https://github.com/artefactual-labs/AIPscan/issues/54

Report of largest files by Storage Service. The tabular report available through the GUI currently lists the 20 largest original files per Storage Service and looks like this:

image

The results can be filtered by file type, so that only original or preservation files are displayed. It defaults to showing files across all types.

The data endpoint underneath as exposed in the API also accepts a `limit` parameter to specify the number of results to return.

I've added tests for the Data component, which confirm that `data.largest_files` is returning data that looks like what we expect, given mocked database query return values. Ideally I'd want to do some integrated tests with a test database as well given how much work the database queries are doing here, but we're currently missing infrastructure for that (whereas Django comes with support for that out-of-the-box, it's something that we'll have to set up on our own using Flask).


peterVG:
Hi @tw4l. Not quite sure on the process for WIP PRs. Presumably you want a review even though it's not finished yet? Will you merge after approval or wait until the code is no longer WIP?

tw4l:
Hey @peterVG, thanks for the code reviews! I was holding off on submitting a proper PR for this until I had a chance to rebase it on the database work and add some tests. I should have it at a point where it's ready for review tomorrow or early next week and flag you when it's ready.

tw4l:
Hi @peterVG, this is now ready for review. Thanks!

peterVG:
* The report works great. Switching between All, Original, or Preservation copies is a really nice touch.
* Changing limit via https://github.com/artefactual-labs/AIPscan/blob/dev/issue-54-largest-files-report/AIPscan/Data/data.py#L245 works great. Would be nice to make this configurable via UI at some point as per inline notes. Do we have an issue for that?
* The tox tests for `data.largest_files` are passing. Thanks for including those and setting an example for how to add tests going forward.
* I like the text-muted class and "small" tag for displaying the UUIDs in an unobtrusive way under the filenames.
* I don't think we should have hyperlinks on the reports. These are intended for print output. I think it looks less elegant when you go to print view to see blue underlined text in the report, especially since it becomes non functional in a PDF or print format. See the tabular format reports for _File format content_ or _AIP contents_ for comparison. Users can click around in the AIP GUIs to jump between AIPs, originals, preservation copies or out to PRONOM.
* I think the "Limit by file type" drop-down box should be excluded from the print output. It's a GUI widget that doesn't belong in a print format.

peterVG (changes requested):
* The report works great. Switching between All, Original, or Preservation copies is a really nice touch.
* Changing limit via https://github.com/artefactual-labs/AIPscan/blob/dev/issue-54-largest-files-report/AIPscan/Data/data.py#L245 works great. Would be nice to make this configurable via UI at some point as per inline notes. Do we have an issue for that?
* The tox tests for `data.largest_files` are passing. Thanks for including those and setting an example for how to add tests going forward.
* I like the text-muted class and "small" tag for displaying the UUIDs in an unobtrusive way under the filenames.
* I don't think we should have hyperlinks on the reports. These are intended for print output. I think it looks less elegant when you go to print view to see blue underlined text in the report, especially since it becomes non functional in a PDF or print format. See the tabular format reports for _File format content_ or _AIP contents_ for comparison. Users can click around in the AIP GUIs to jump between AIPs, originals, preservation copies or out to PRONOM.
* I think the "Limit by file type" drop-down box should be excluded from the print output. It's a GUI widget that doesn't belong in a print format.

tw4l:
Hi @peterVG - flagging this for re-review.

I think I've addressed your comments in the follow-up commits. The changes made are:

1. Removing hyperlinks from the report (I also added the AIP UUID, since we no longer have a link). This is what the report looks like now:

image

2. Removing the file type selector from the print view. 

3. Adding a functional "limit" parameter to the URL, to take advantage of how it's configurable in the view. I stopped short of adding a GUI component in the report itself, but if you add "&limit=5" (or some other number) to the end of the largest files report URL, it will take effect and will be respected when changing file types. I've also made an issue for adding a GUI component to control this parameter: https://github.com/artefactual-labs/AIPscan/issues/77

4. A few small UI tweaks (renaming "Limit by file type" to just "File type", slight modifications to the report header for the sake of readability).

Let me know if you'd like to see further changes!

peterVG (approved):
1. Confirmed. Thank you. Recognizing that there's still an open discussion about showing this report/info "somewhere" in AIPscan is still desirable.
2. Confirmed. Also tested that the print view corresponds to the "All", "Original files", or "Preservation Files" options. Now that the drop-down is gone. I wonder if we should add that info to the print report header. Now if you have one of these printed reports laying around on your desk for a few days it may not be clear if you are looking at a list of "All files", "Original files", or "Preservation Files". But I don't think that should hold up this PR. Should we add an issue for that?
3. Cool, this works great. Nice touch. Confirmed that limit "sticks" through file type changes.
4. Looks clean and tidy. Ready to ship out!

tw4l:
Thanks @peterVG!

> Also tested that the print view corresponds to the "All", "Original files", or "Preservation Files" options. Now that the drop-down is gone, I wonder if we should add that info to the print report header. Now if you have one of these printed reports laying around on your desk for a few days it may not be clear if you are looking at a list of "All files", "Original files", or "Preservation Files"

I'll add file type to the report template header before merging - should be a quick addition.

Issue #18, opened 2020-06-23
Populate Agents table


peterVG:
Requires m:m relationship with Events table

PR #65, opened 2020-09-24
Agents-transfers logging reports

ross-spencer:
This finally completes an API endpoint and basic reports to list information about users and ingests for which they were responsible for. Delivering a tabular representation of that work and a prototype timeline using plotly express. 

The API endpoint to access the data will be at: 

* `curl -X GET "http://127.0.0.1:5000/api/data/agents-transfers/1" -H "accept: application/json"`

**Prototype gantt**

image

**Table**

image

The work leverages much of the testing efforts of Tessa so thank you for that! And also tries to comply with new efforts around base templates as well. Hopefully it's looking okay. 

@peterVG we'd spoken before about what methods we might use to deliver charts and there might be some good exploration to do there. For this work the code that enables that is located in this commit which you might be interested in: https://github.com/artefactual-labs/AIPscan/pull/65/commits/d8cf25def1b92d2339eeeeabe735e10125f8705a 

Connected to #18


AIPscan/Reporter/tests/test_ingest_log.py
```diff
@@ -0,0 +1,106 @@
+# -*- coding: utf-8 -*-
+
+"""Tests for the agents transfer log."""
```
tw4l, line 3: These tests look great - bravo! :clap:

AIPscan/Data/data.py
```diff
@@ -422,3 +437,44 @@ def aips_by_puid(storage_service_id, puid, original_files=True):
...
+    for aip in aips:
+        files = File.query.filter_by(aip_id=aip.id)
+        for file in files[:1]:
+            events = Event.query.filter_by(file_id=file.id, type=EVENT_TYPE).all()
+            for event in events:
+                log_line = {}
+                log_line[FIELD_AIP] = aip.uuid
+                log_line[FIELD_AIP_NAME] = aip.transfer_name
+                log_line[FIELD_INGEST_START_DATE] = str(event.date)
+                log_line[FIELD_INGEST_FINISH_DATE] = str(aip.create_date)
```
tw4l, line 474: This is interesting. It makes perfect sense as implemented. But it makes me wonder about which of these values we should be using for the start and end date parameters on our reports, and how we communicate that choice.

For this particular report, I imagine we'd want to use the ingestion `Event` date for start date, and `AIP.create_date` for end date, to keep it consistent.

For other reports, I've been using `AIP.create_date` for both dates in my WIP code (rationale being that it reflects what's in the Storage Service/Archival Storage tab) but it looks like we've used the ingestion `Event` date for both elsewhere and I can understand why that's also a reasonable choice. Granted, in most cases they'll be the same since those parameters only accept YYYY-MM-DD values, but they might not be for some very long-running ingests. What do you think?

tw4l (commented):
Hi @ross-spencer. This is looking great - nice work! I've left comments where I had questions or think there are some opportunities to tidy bits of the code up a bit, but overall it's already in good shape.

I'm wondering what the Gantt chart will look like for Storage Services that have been in production for months or years. Would the individual ingests still be visible and hoverable? It seems like the built-in plotly toolbar's zoom function would help but I suspect that might point to the need to add support for start/end date parameters in the near future. That said, I think it would be reasonable to do so in a follow-up PR, and as I pointed out offline my next PR will contain some helpers that might make that a bit simpler.

AIPscan/Reporter/templates/report_ingest_log_tabular.html
```diff
@@ -0,0 +1,71 @@
...
+{% endblock %}
+
+{% block content %}
+
+  <div class="alert alert-secondary">
+    <a class="noprint" onClick="window.print();return false">
+      <button type="button" class="btn btn-info" style="float:right;">Print</button>
+    </a>
+    <strong>User ingest log</strong>
+    </br>
```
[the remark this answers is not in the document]
ross-spencer: I had this waaaay wrong anyway. The XHTML version that can be used is `<br/>`.

AIPscan/Data/data.py
```diff
@@ -422,3 +437,44 @@ def aips_by_puid(storage_service_id, puid, original_files=True):
...
+        return report
+
+    aips = AIP.query.filter_by(storage_service_id=storage_service.id).all()
+
+    EVENT_TYPE = "ingestion"
+    AGENT_TYPE = "Archivematica user"
+
+    for aip in aips:
+        files = File.query.filter_by(aip_id=aip.id)
+        for file in files[:1]:
```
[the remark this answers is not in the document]
ross-spencer: Nice!

AIPscan/Data/tests/test_agents_transfers.py
```diff
@@ -0,0 +1,51 @@
...
+            1,
+            "111111111111-1111-1111-11111111",
+            "test aip",
+            "user one",
+        ),
+        # Request for a non-existent storage service.
+        (30, None, 0, "", "", ""),
+    ],
+)
+def test_aips_by_file_format_aip_elements(
```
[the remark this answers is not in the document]
ross-spencer: Whoops! Busted trying to get my tests to work! Will do!

AIPscan/Data/data.py
ross-spencer: I wonder if you can clarify that comment/question a little Tessa? Primarily: 

> For this particular report, I imagine we'd want to use the ingestion Event date for start date, and AIP.create_date for end date, to keep it consistent.

Which I think I am doing?

AIPscan/Reporter/templates/report_ingest_log_tabular.html
```diff
@@ -0,0 +1,71 @@
+<!-- Ingest log tabular report. -->
+
+{% extends "report_base.html" %}
+
+{% block extra_css %}
+  <link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='css/style_ingest_log.css') }}">
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.1/css/all.min.css"
```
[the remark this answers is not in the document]
ross-spencer: I was anticipating the question. Have you more background on that Tessa - I have seen this suggested elsewhere but not without any particular rationale being clear. What I could half fathom was that the amount of load on production servers generated made it desirable, but I don't imagine AIPscan being used that heavily. I also believe network access to the outside world will be a requirement in the long-run with the potential requirement to access external data sources. 

But anyway, that's my understanding, some background on the why would be beneficial for me? 

I ended up not bringing it into static because it ultimately feels harder to maintain over time. And this was the pragmatic approach with very little over-head to try, i.e. this is the first report using the icons for a bit of flare so I didn't know if it would stick. That being said, I haven't a strong opinion. Though I'd like to question the norm.

AIPscan/Data/data.py
tw4l: Yes! To be clear I don't think anything in this PR needs to change.

I meant this more as a conversation prompt for AIPscan as a whole, looking ahead to adding `start_date` and `end_date` parameters for more reports in the reports UI/API, including potentially this one.

As I said above, I think how you've implemented the dates for this report makes perfect sense, and it would make sense to continue by applying the same logic to date inputs that come in with the request from the GUI or API. But that would mean that those parameters would be a little different to how they are implemented in the "File format countt" (where the ingestion `Event` date is used for both `start_date` and `end_dates`), which is also different to how I had been thinking about it in relation to the "Format version count" report I'm about to submit a PR for (where I've using `AIP.create_date`, as the date that the user would see in their Archival Storage tab/Storage Service and might want to carry over into AIPscan).

It's more a question than a prompt for change, and maybe we can hash it out in other PRs or conversations, but I thought that since your work prompted the thought and it potentially has an impact on the work I'm just about to submit, I'd check if you had any initial thoughts.

AIPscan/Reporter/templates/report_ingest_log_tabular.html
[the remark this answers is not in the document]
tw4l: Hi Ross. FWIW I'm also not convinced of the load argument (the browser's going to have to pull that file down from somewhere to cache it, and it's probably faster some or a lot of the time doing that from a a local CDN cache near them than from our server).

I had two angles on this:

One is that it seems like we might already have (a slightly older version of) this file in our static directory: https://github.com/artefactual-labs/AIPscan/blob/main/AIPscan/static/css/fontawesome.css.

But primarily I was thinking of the issue of network access to the outside world. It's possible that I'm being too conservative here because it's been an issue for Brunnhilde in the not-too-distant past (which at one point download Bootstrap assets from GitHub until it became clear that was an issue for folks), but I guess I wouldn't be too surprised if some of our clients would prefer not to have outside network access on their Archivematica/AIPscan servers unless absolutely necessary, and from what I can tell this is the first thing in the codebase that would complicate that.

> I also believe network access to the outside world will be a requirement in the long-run with the potential requirement to access external data sources.

That's interesting, and I suspect you're ahead of me in your thinking on that.

> I ended up not bringing it into static because it ultimately feels harder to maintain over time.

I can see this as well, although I wonder if there's any reason to question in a long-term scenario whether assets may at some point no longer be available via the CDN at some point?

Honestly, this isn't a sticking point for me, so if you'd rather leave it as-is and change it down the road if someone complains, I think that would be okay.

AIPscan/Data/data.py
tw4l: With the same caveat as above that I don't intend these as a blocker for this PR, I guess another way to think about this might be - if a transfer is created in January and stored in the transfer backlog, and then that transfer (or parts of it, or parts of several transfers) get run through ingest and turned into an AIP in June, the ingestion event and the AIP creation date will be very different. When we say "start date" and "end date" in relation to something like the "File format count" report, do we mean when the file was first ingested into an Archivematica instance, or when it became part of an AIP? Might that answer change for other reports? Is there a way to communicate what it means to the user?

tw4l: @ross-spencer I'm marking this as resolved since it shouldn't get in your way, and we can have the conversation in the PR I'm just about to submit: https://github.com/artefactual-labs/AIPscan/pull/98

AIPscan/Reporter/report_ingest_log.py
```diff
@@ -0,0 +1,67 @@
...
+    """Remove microseconds from the given date string."""
+    return str(date_string).split(".")[0]
+
+
+def _get_datetime(date_string):
+    """Format the supplied date string to remove the seconds for
+    compatibility with the GANTT report.
+    """
+    DATE_FORMAT_FULL = "%Y-%m-%d %H:%M:%S"
+    return datetime.strptime(_split_ms(date_string), DATE_FORMAT_FULL)
```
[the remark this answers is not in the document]
ross-spencer: I've given it a try. I do like them and suspect they'll be used again. 

NB. Incidentally you'll see a similar one in the test module now, but I have purposely used something different so the tests aren't using the same code that it's trying to prove.

AIPscan/Data/data.py
ross-spencer: Oh cool. Yeah, these are all good points. Initial instinct at a low level is to encode some of this in helpers in AIPscan at least that describe the heuristic and values being retrieved more completely than I've done here (that will also promote greater accuracy). On a higher-level Joel recognized early on that we might want a domain model for this piece. (Well, all parts of our work really). But this is a good early catch, thanks for putting in the forefront of my mind too.

AIPscan/Reporter/templates/report_ingest_log_tabular.html
```diff
@@ -0,0 +1,71 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {% for row in data %}
+            <tr>
+              <td id="user">
+                <b>{{ row["User"] }}</b>
+              </td>
+              <td>
+                {{ row["AIPName"] }} <code>({{ row["AIP"] }})</code>
```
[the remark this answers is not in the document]
ross-spencer: Good spot, and have also implemented most of that! I like the separate column too so have done that for AIP UUID. :+1:

AIPscan/Reporter/templates/report_ingest_log_tabular.html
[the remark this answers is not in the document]
ross-spencer: Thanks for the response Tessa, I'm glad I'm not quite alone in some of those thoughts. Heeding your past experiences i've tried to import it static, and it looks like it is working okay - I've wrestled with many white squares! But let me know.

ross-spencer:
Thanks Tessa - resubmitted for code review. Also noted offline, we're thinking we can improve the date handling around the plot chart sometime down the line using some of your new handlers. I'm looking forward to review those, and thanks again for the conversation around this.

AIPscan/Data/data.py
```diff
@@ -463,8 +463,8 @@ def agents_transfers(storage_service_id):
     AGENT_TYPE = "Archivematica user"
 
     for aip in aips:
-        files = File.query.filter_by(aip_id=aip.id)
-        for file in files[:1]:
+        file = File.query.filter_by(aip_id=aip.id).first()
```
tw4l, line 451: Good catch on `.first()` being the right choice here - looks great

AIPscan/Reporter/templates/report_ingest_log_tabular.html
```diff
@@ -51,7 +54,10 @@
                 <b>{{ row["User"] }}</b>
               </td>
               <td>
-                {{ row["AIPName"] }} <code>({{ row["AIP"] }})</code>
+                {{ row["AIPName"] }}
+              </td>
+              <td>
+                <code><span class="text-muted">{{ row["AIPUUID"] }}</span></code>
```
tw4l, line 57: With the AIP UUID is in it's own column, I think just `<td>{{ row["AIPUUID"] }}</td>` might suffice here. In the other tables, we don't have the values in that column muted.

ross-spencer: This is just a stylistic decision on my part here so I appreciate it might look redundant? 

image

I kinda like it. It's one way of adding flare. But can change.

ross-spencer:
@tw4l there are a couple of commits on top of the first push here now - I noticed a broken endpoint (i broke it), and some documentation issues that it felt appropriate to fix.

tw4l:
@ross-spencer cool, thanks! I'm planning to finish up my review tomorrow morning.

AIPscan/Data/data.py
```diff
@@ -6,7 +6,6 @@
 
 VALID_FILE_TYPES = set(item.value for item in FileType)
 
-FIELD_AIP = "AIP"
```
tw4l, line 10: I think this is still in use in a couple places in the codebase that would need to be changed before we can safely remove it:
* https://github.com/artefactual-labs/AIPscan/blob/91a8237998e0edb4ce79a4419878f92a21fe9908/AIPscan/Reporter/report_largest_files.py#L29
* https://github.com/artefactual-labs/AIPscan/blob/91a8237998e0edb4ce79a4419878f92a21fe9908/AIPscan/Reporter/helpers.py#L31

I think those were both my doing, so mea culpa! Feel free to change those to `data.FIELD_AIP_NAME` and delete this, or just leave `FIELD_AIP` in for now and I'll clean it up, as you prefer.

AIPscan/Reporter/templates/report_ingest_log_tabular.html
tw4l: Yeah, fair enough! There should be room for some design choices here :smile:

AIPscan/API/namespace_data.py
```diff
@@ -42,7 +42,7 @@ class FMTList(Resource):
         },
     )
     def get(self, storage_service_id):
-        """AIP overview One"""
+        """List PUIDs and AIPs containing examples of them"""
```
tw4l, line 45: These are great!

tw4l:
Hi @ross-spencer - this is looking great and I think we're very nearly there. There are just two areas I think could use a little attention:
1. Linting seems to be failing for `AIPscan/Data/tests/test_agents_transfers.py` when I run the tests via `tox`. I think it just needs `black` to be run on it.
2. Thank you for bringing the font-awesome assets in locally! There are a lot more files than I was expecting, and it's making it a bit hard to pick out our own JS and CSS files from the vendor ones. Would you mind seeing if we can group them together, maybe in a `vendor/fontawesome` subdirectory, or something like that? I recognize that we haven't done that for vendor assets yet, but it seems like this might be a good place to start with that. I'd also encourage seeing if we need all of these assets or whether a subset would suffice, but I'd understand if at this stage you'd rather put that time and effort elsewhere so please don't see that as a hard requirement.

AIPscan/Reporter/templates/report_ingest_log_tabular.html
ross-spencer: In general too, I think it's nice to have fixed width strings like a UUID in a fixed-width font for in a big enough list their alignment starts to get messy quite quickly. There would clearly be other ways to do this, but `<code>` is quite convenient.

AIPscan/Data/data.py
ross-spencer: I also only counted those two so have changed them and tried the largest file report and it looks okay...  so :crossed_fingers: we got that.

ross-spencer:
Should be good now @tw4l lmk!

ross-spencer:
EDIT: You got me thinking about dates more and I spotted an issue with the query as a result so there are two other commits. They work like this:

1) Creates a failure condition in the tests.
2) Writes the fix to correct that. 

Now I'm hoping this PR is looking good! :crossed_fingers:

AIPscan/Reporter/templates/report_ingest_log_tabular.html
tw4l: I was puzzling a bit at the use of the `<code>` block before, but this makes sense!

tw4l (approved):
Tests and linting are passing, code looks good, all issues resolved. LGTM! :+1: Good work too on updating the query and tests!

image

Issue #68, opened 2020-10-05
AIP object creation fails due to missing field name


tw4l:
The `uuid` field name is missing from the AIP constructor in the Aggregator module's `create_aip_object` database helper: https://github.com/artefactual-labs/AIPscan/blob/591a773f49e0f301f6ebea64141e085105d91379/AIPscan/Aggregator/database_helpers.py#L143.

This is periodically preventing AIPs from being written correctly to the database.

PR #69, opened 2020-10-05
Add missing uuid field name in AIP constructor

tw4l:
Fixes https://github.com/artefactual-labs/AIPscan/issues/68


peterVG:
I just ran a Fetch Job with the pre-fix code and AIP uuids were written to the database correctly. Weird that it works sometimes and not other times? It seems like it should always fail. Anyway makes sense to fix and avoid the question altogether.

I re-ran using the fix on `dev-issue-68-aip-uuid` and all AIP uuids were correctly inserted into the db. So LGTM!

peterVG (approved):

Issue #62, opened 2020-09-23
Files models don't contain a field for filepath


tw4l:
The models that store information on files (currently `originals` and `copies`), store filenames but not filepaths. I would like to be able to include filepaths in some reports so that users can more quickly identify where particular files are within their AIPs, and because in the case of complex digital objects the paths can be semantically meaningful.

Related to #42, #55

tw4l:
Field added in https://github.com/artefactual-labs/AIPscan/pull/71

Issue #60, opened 2020-09-22
Model class names should be capitalized and singular


tw4l:
AIPscan's model classes are currently all lowercase: https://github.com/artefactual-labs/AIPscan/blob/da8fc4c939a927cf0dbd54d7ab2fedc6f625e140/AIPscan/models.py

Per [PEP8](https://www.python.org/dev/peps/pep-0008/#class-names), class names should use the CapWords convention: 

Flask's own [style guide](https://flask.palletsprojects.com/en/1.1.x/styleguide/#naming-conventions) suggests the same, with the added suggestion that "acronyms kept uppercase (`HTTPWriter` and not `HttpWriter`)".

By convention, the class names should also be singular, so for example `storage_services` should be `StorageService`, `aips` should be `AIP` and so on.

This will make it easier to identify Models in the code base, e.g. when they are used in SQLAlchemy queries.

tw4l:
Related to https://github.com/artefactual-labs/AIPscan/issues/55

ross-spencer:
It makes it easier to find names for variables where, as an example, `agent` can happily be my instance of `Agent`.

Issue #42, opened 2020-07-28
There isn't an easy to access entity relationship diagram (ERD)


ross-spencer:
We haven't a ERD for AIPscan. We can access the schema from `sqlite` but something ERD-like might also help developers.
```
sqlite> .schema --indent *
CREATE TABLE storage_services(
  id INTEGER NOT NULL,
  name VARCHAR(255),
  url VARCHAR(255),
  user_name VARCHAR(255),
  api_key VARCHAR(255),
  download_limit INTEGER,
  download_offset INTEGER,
  "default" BOOLEAN,
  PRIMARY KEY(id),
  CHECK("default" IN(0, 1))
);
CREATE UNIQUE INDEX ix_storage_services_name ON storage_services(name);
CREATE TABLE agents(
  id INTEGER NOT NULL,
  type VARCHAR(255),
  value VARCHAR(255),
  PRIMARY KEY(id)
);
CREATE INDEX ix_agents_value ON agents(value);
CREATE INDEX ix_agents_type ON agents(type);
CREATE TABLE fetch_jobs(
  id INTEGER NOT NULL,
  total_packages INTEGER,
  total_aips INTEGER,
  total_deleted_aips INTEGER,
  download_start DATETIME,
  download_end DATETIME,
  download_directory VARCHAR(255),
  storage_service_id INTEGER NOT NULL,
  PRIMARY KEY(id),
  FOREIGN KEY(storage_service_id) REFERENCES storage_services(id)
);
CREATE TABLE aips(
  id INTEGER NOT NULL,
  uuid VARCHAR(255),
  transfer_name VARCHAR(255),
  create_date DATETIME,
  originals_count INTEGER,
  copies_count INTEGER,
  storage_service_id INTEGER NOT NULL,
  fetch_job_id INTEGER NOT NULL,
  PRIMARY KEY(id),
  FOREIGN KEY(storage_service_id) REFERENCES storage_services(id),
  FOREIGN KEY(fetch_job_id) REFERENCES fetch_jobs(id)
);
CREATE INDEX ix_aips_uuid ON aips(uuid);
CREATE TABLE originals(
  id INTEGER NOT NULL,
  name VARCHAR(255),
  uuid VARCHAR(255),
  size INTEGER,
  puid VARCHAR(255),
  format VARCHAR(255),
  format_version VARCHAR(255),
  checksum_type VARCHAR(255),
  checksum_value VARCHAR(255),
  related_uuid VARCHAR(255),
[...]

Issue #49, opened 2020-08-17
Add database migrations


peterVG:

aseles13:
This issue does not have an explanation and based on the approve merge request #71 the work related to this issue was completed

PR #71, opened 2020-10-05
Refactor database models

tw4l:
Refactors database models to conform to PEP8/Python style conventions and to use a single self-referential `File` model for all types of digital objects.

Connects to #55 , #60 , and #62.

Some notes:

## Celery models

You'll note that I haven't touched the Celery models. I think the Celery tasks and how we're keeping track of their status deserves its own look. Judging from my initial look that might take a little time and consideration, so I thought it best to address separately, outside of the scope of this PR.

## File properties

Per our last conversation @peterVG, this work means that for any given file it is possible to access its original file or preservation derivative directly from the model object.

So if we have an original file, e.g.:

```
original_file = File.query.get(some_id)
```

`original_file.derivatives` is a list of all `File` objects which link back to this as their original file via `File.original_file_id`. Currently this list will be empty if there is no linked preservation file, or will consist of a single preservation file if one exists. Moving forward, this could accommodate additional types of derivatives as well (e.g. Dataverse derivatives) if we want to track those in the application.

And if we have a preservation file which we've accessed via `original_file.derivatives[0]` or a query such as:

```
preservation_file = File.query.filter_by(
    file_type=FileType.preservation, original_file_id=original_file.id
).first()
```

we can get access the original file via `preservation_file.original_file`.

## Other notes

I've tried to do some style clean-up as I went as well (e.g. converting camelCase variables to snake_case and using `filesizeformat` in templates to format human-readable sizes rather than using our own helper in views) where I didn't have to go too far out of my way to do it.

All existing tests are passing after the refactor, but I did notice that tests in the `Aggregator` component are rather sparse at the moment. We should plan to focus on those in an upcoming portion of the work to ensure that the data being used for reporting is accurate.

We will need to update the entity relationship diagram as per #42 as well. This is a pretty significant refactoring of the database models. Given the early stage of development and the fact that we don't have deployments of AIPscan yet, I didn't attempt to put together migrations. That might be another area we want to put some attention into in an upcoming sprint, and hopefully this gives us a good base to work from moving forward.


peterVG:
Thank you for the PEP cleanup! :-)

I really like the forward-compatible functionality this refactor give us. It's an elegant solution. The `original_file.derivatives` and `preservation_file.original_file` getters are brilliant.

I agree that we can get away without worrying about migrations for a while longer. Really until we have a release where we know actual users are storing actual data in AIPscan. We have a placeholder issue for this: https://github.com/artefactual-labs/AIPscan/issues/49

I rebuilt a new AIPscan db using the data model on this branch and ran a fresh Fetch Job. The `file_type` and `original_file_id` columns were populated correctly. All AIP and File view templates are working correctly as are the file format reports. So LGTM!

peterVG (approved):

Issue #81, opened 2020-10-21
Resolve circular imports to get AIPscan/__init__.py to pass linting


tw4l:
Related to #35. See this comment particularly: https://github.com/artefactual-labs/AIPscan/issues/35#issuecomment-693650138

From cursory reading so far, this seems to be best resolved by following the [Application Factories](https://flask.palletsprojects.com/en/1.1.x/tutorial/factory/) Flask pattern. This should also help with #79.

ross-spencer:
@tw4l as this is now a separate issue maybe we can also close out #35 some-how or get some commentary on that? Perhaps we can take the temporary approach I describe in the comment there and disable the warnings around the circular imports - that way we would have linting on and ready to go so we can immediately take advantage of that - and then this issue sits as a discrete new task. What do you think? @peterVG, thoughts too?

tw4l:
@ross-spencer Agreed, I think the temporary approach you described would be a a good way forward so that we still get the benefits of #35, so I'm in favour.

peterVG:
Also related to #31

tw4l:
Resolved in https://github.com/artefactual-labs/AIPscan/pull/88

Issue #79, opened 2020-10-21
Add app config and test database for integration tests


tw4l:
In many of our newer reports such as those in #75, most of the business logic is in somewhat complicated SQLAlchemy queries, so it is difficult to get appropriate test coverage (to ensure that the queries are returning the data we want, and to avoid future regressions) without integration tests that create state in a test database.

This would also be helpful in the `Aggregator` component, to ensure that METS files are being parsed to the database as expected.

Relevant bits of Flask documentation:

* [Testing Flask Applications](https://flask.palletsprojects.com/en/1.1.x/testing/)
* [Configuration Handling](https://flask.palletsprojects.com/en/1.1.x/config/)
* [Application Factories](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/)

tw4l:
Added in https://github.com/artefactual-labs/AIPscan/pull/88

PR #88, opened 2020-11-02
Enable integration testing via Flask Application Factory pattern

tw4l:
Connected to #79, #81, #35 

This PR introduces the [Application Factory](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/) pattern to AIPscan, which resolves our linting errors, enables multiple application configurations (e.g. for production, development, and testing), and  provides us the ability to use an in-memory test database for tests that require accessing database state. It also includes some new test fixtures (see more below) and an example integration test.

Along the way, this work also had the following consequences:
* Changes to how Celery is configured:
    * Celery is now initialized in `AIPscan.extensions` prior to being provided Flask application context in `AIPscan.celery.configure_celery`.
    * The configuration details for the Celery result backend and broker were moved to their own config file and are now environment-aware.
    * The Celery worker should now be started with: `celery worker -A AIPscan.worker.celery --loglevel=info` (the instructions in the README have also been updated)
* The `db.create_all()` call in the `Aggregator` Blueprint's `__init__` module had to be removed, so the database must be explicitly created prior to running the application.
* It was necessary to remove the `@app.route("/")` handler from the `Aggregator` component. To avoid any user-facing changes, I created a very small `Home` blueprint to add a route for the root ("/") path which simply redirects to the `Aggregator` view function `@app.route("/")` was previously decorating.
* Tests that invoke Flask or SQLAlchemy-related code now need to be aware of their [application context](https://flask-sqlalchemy.palletsprojects.com/en/2.x/contexts/). The `app_instance` fixture (added in this work to the `Aggregator` and `Data` components) gives a demonstration of how to provide this application context without requiring changes to our tests (edited to account for latest commit).

Ideally we would want to define fixtures such as `app_instance` once rather than separately in each Blueprint, but I think that would require co-locating our tests within a single module (e.g. `AIPscan.tests`) in order to use fixtures defined in a `conftest` module in tests for any of our Blueprints. If we did go that way, we might keep the per-Blueprint structure via sub-modules within `AIPscan.tests`. I'd like some input on whether that sounds reasonable, or if there might be other consequences that I'm not seeing.

And perhaps the nicest consequence:

image


AIPscan/application.py
```diff
@@ -0,0 +1,31 @@
...
+    app.config.from_object(CONFIGS[config_name])
+
+    app.register_blueprint(aggregator, url_prefix="/aggregator")
+    app.register_blueprint(reporter, url_prefix="/reporter")
+    app.register_blueprint(user, url_prefix="/user")
+    app.register_blueprint(api)
+    app.register_blueprint(home)
+
+    db.init_app(app)
+    configure_celery(app)
```
ross-spencer, line 29: From a discussion I think at least one change we'll make is move these lines into a context manager per: https://gist.github.com/ross-spencer/3bfa40babca5b6553ee797d36c548d3e as this then means the database tables can be created with an application context, i.e. Flask doesn't complain.

AIPscan/Aggregator/tests/conftest.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
ross-spencer, line 1: Question noted in the PR comment: `Ideally we would want to define fixtures such as app_instance once rather than separately in each Blueprint,` yeah, it does seem like an unnecessary redundancy but I'm comfortable with it for now - especially if it means not co-locating the tests. Also, suggested solution noted. IDK myself yet. Let's have a think. This work is definitely two steps forward so that's :ok_hand: 

One thing I would love to see modified slightly is the naming of this `conftest.py` and the other one. Primarily as it isn't a test, and secondly because it loses its snake casing so it's a bit of a portmanteau of sorts. Off the top off my head - `app_fixtures.py`?

ross-spencer (commented):
image

fist-pump

Great consequence. 

I was going to make the small changes required to this piece but I have also been short of time today, and want to get to https://github.com/artefactual-labs/AIPscan/pull/89 which is another reason I've made review comments here instead and then will move onto that. 

The review comments here I think are just small, largely stylistic. 

RE: https://github.com/artefactual-labs/AIPscan/pull/89 I mentioned to Peter it seems logical to me to merge that first, then rebase this one on top of that. Which is the part I haven't had time for at 20:00 tonight with a couple of other things to look at. I'll leave it up to you two to discuss. 

Nice work Tessa.

run.py
```diff
@@ -1,4 +1,9 @@
...
+import os
+
+from AIPscan.application import create_app
+
 
 if __name__ == "__main__":
-    app.run(debug=True, host="0.0.0.0")
+    config_name = os.environ.get("FLASK_CONFIG", "default")
+    app = create_app(config_name)
+    app.run(host="0.0.0.0")
```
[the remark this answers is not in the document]
tw4l: Makes sense to me!

AIPscan/Aggregator/tests/conftest.py
tw4l: > One thing I would love to see modified slightly is the naming of this conftest.py and the other one.

I fully agree and would love to give it a better name, but unfortunately the name is [built-in to pytest](https://docs.pytest.org/en/stable/fixture.html?highlight=conftest#conftest-py-sharing-fixture-functions) and the fixture won't be auto-discovered if we change the module name.

>  yeah, it does seem like an unnecessary redundancy but I'm comfortable with it for now - especially if it means not co-locating the tests. Also, suggested solution noted. IDK myself yet. Let's have a think.

After going through the pytest documentation again, I think it may actually be possible to remove the duplication by moving the `conftest.py` file into the main `AIPscan` package. Will test that out and see if it gives us another potential solution that doesn't require moving the tests around...

ross-spencer: > I fully agree and would love to give it a better name, but unfortunately the name is built-in to pytest and the fixture won't be auto-discovered if we change the module name.

TIL!!

tw4l: It was new to me too! The top answer on this SO post gives a good overview of all the things that can be done with `conftest.py`: https://stackoverflow.com/questions/34466027/in-pytest-what-is-the-use-of-conftest-py-files

AIPscan/worker.py
```diff
@@ -0,0 +1,5 @@
+from AIPscan.application import create_app
+from AIPscan.celery import configure_celery
+
```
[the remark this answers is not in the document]
tw4l: Done and done!

I do wonder if we might want to remove the `# -*- coding: utf-8 -*-` headers eventually, since we don't intend (at least as far as I know) to support Python 2? That said, they're not doing any harm and I do see the value in consistency so maybe nbd.

create_aipscan.db.py
```diff
@@ -4,5 +4,8 @@
 # ./create_aipscan.db.py
 
 from AIPscan import db
+from AIPscan.application import create_app
 
+app = create_app()
+app.app_context().push()
 db.create_all()
```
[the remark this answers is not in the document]
tw4l: Removed in https://github.com/artefactual-labs/AIPscan/pull/88/commits/125f1ac8cdf66b35cc4cd65f0a3053e510ee30fe

tw4l:
Hi @ross-spencer - I've made all the changes I think need to be made based on the review comments. That includes reverting to automatically creating the database, which I'm glad you pushed for - it is a nicer experience this way, and moving our `conftest.py` up into the `AIPscan` package so that our `app_instance` pytest fixture is available across all of the blueprints.

In addition, I've moved the application factory into `AIPscan.__init__` (largely just so it's where other developers with Flask experience will expect it and to reduce the number of modules, though there may be other advantages as noted in the commit message) and moved the blueprint imports into the application factory function.

Tests and linting are all still passing following the changes, so if you're happy I think all that remains is to rebase on #89 after it's merged.

ross-spencer (approved):
Superb work. Thanks a lot Tessa. Awaiting PVG, but no reason not to approve. Feel free to ping him or I if you feel a once-over on any changes following that is necessary.

Issue #93, opened 2020-11-26
Add "Format version count" report


tw4l:
The AIPscan reports UI includes a stub for a "Format version count" report (report 3) that  is not yet implemented. Currently, when a user clicks on this report, a "Coming soon" message is flashed instead of generating a report. We should implement this report and remove the "Coming soon" alert.

Issue #83, opened 2020-10-21
Handle AttributeErrors when reading file properties to database more gracefully


tw4l:
Currently if there is an attribute error at any point during the parsing of a file's properties from a METS `FSEntry` into the AIPscan database, the file gets written to the database as an ISO disk image, which is likely to be misleading: https://github.com/artefactual-labs/AIPscan/blob/a25d742f424e8800fee47b9b1eef9b84111fb0b1/AIPscan/Aggregator/database_helpers.py#L196-L200.

ross-spencer:
There is a nod to this particular issue here: https://github.com/artefactual-labs/AIPscan/issues/47#issuecomment-691310678 as I hoped to leave space for the developer working on #47 to consider this. So connected to #47, but a valid assertion. I should also have been clearer that the root cause of this is in `metsrw` https://github.com/archivematica/Issues/issues/1277

tw4l:
Thanks @ross-spencer, I missed the connection to #47. That's good context!

PR #98, opened 2020-11-30
Add format versions count report

tw4l:
Connected to #93 

This PR implements the "Format versions count" report in the UI and API, with support for limiting results by start and end date. This commit also removes the Bootstrap alerts from the report selection template, as there is no longer the need for any "Coming soon"-type messages, and users should be able to tell that their reports are loading in a new tab without these alerts.

**NB:** With reference to this conversation in https://github.com/artefactual-labs/AIPscan/pull/65#discussion_r534446391, I suspect that some changes will need to be made around implementing the `start_date` and `end_date` parameters before merging. This might mean changing this report to use files' ingestion `Event` date or it might mean modifying the "File format count" report to use `AIP.create_date`, as I've done here. My rationale for using `AIP.create_date` is that that date is likely the most visible date for Archivematica users, since it is displayed in both the Archival Storage tab and the Storage Service's Package tab. But honestly, I'm not really sure which is the better approach and would like some input, as I'd like to set a standard practice and use that moving forward wherever possible.

In many cases, file ingestion dates and AIP creation dates will be the same, or at least similar. But in some cases, as in the case of files that are stored in a transfer backlog for weeks or months prior to becoming part of a SIP (and, eventually, an AIP), they may end up being quite different.

So when we say "start date" and "end date" in relation to something like this report, or the "File format count" report, do we mean when the file was first ingested into an Archivematica instance, or when it became part of an AIP? Might that answer change for other reports? How might we better communicate what these dates mean to users? (I've tried in the API documentation with some success, I think, but am more at a loss for what to do in the GUI)

### DIFF
--- a/AIPscan/API/namespace_report_data.py
+++ b/AIPscan/API/namespace_report_data.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+
+"""Report data namespace
+
+Return data optimized for reports implemented in the Reporter
+blueprint for consumption by external systems.
+"""
+
+from flask import request
+from flask_restx import Namespace, Resource
+
+from AIPscan.helpers import parse_bool, parse_datetime_bound
+from AIPscan.API import fields
+from AIPscan.Data import report_data
+
+api = Namespace(
+    "report-data", description="Retrieve data optimized for AIPscan reports"
+)
+
+
+@api.route("/format-version-count/<storage_service_id>")
+class FormatVersionList(Resource):
+    @api.doc(
+        "list_format_versions",
+        params={
+            fields.FIELD_START_DATE: {
+                "description": "AIP creation start date (inclusive, YYYY-MM-DD)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_END_DATE: {
+                "description": "AIP creation end date (inclusive, YYYY-MM-DD)",
+                "in": "query",
+                "type": "str",
+            },
+        },
+    )
+    def get(self, storage_service_id):
+        """List file format versions with file count and size"""
+        start_date = parse_datetime_bound(request.args.get(fields.FIELD_START_DATE))
+        end_date = parse_datetime_bound(
+            request.args.get(fields.FIELD_END_DATE), upper=True
+        )
+
+        return report_data.format_versions_count(
+            storage_service_id=storage_service_id,
+            start_date=start_date,
+            end_date=end_date,
+        )
+
+
+@api.route("/largest-files/<storage_service_id>")
+class LargestFileList(Resource):
+    @api.doc(
+        "list_formats",
+        params={
+            fields.FIELD_FILE_TYPE: {
+                "description": "Optional file type filter (original or preservation)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_LIMIT: {
+                "description": "Number of results to return (default is 20)",
+                "in": "query",
+                "type": "int",
+            },
+        },
+    )
+    def get(self, storage_service_id, file_type=None, limit=20):
+        """List largest files"""
+        file_type = request.args.get(fields.FIELD_FILE_TYPE)
+        try:
+            limit = int(request.args.get(fields.FIELD_LIMIT, 20))
+        except ValueError:
+            pass
+        return report_data.largest_files(
+            storage_service_id=storage_service_id, file_type=file_type, limit=limit
+        )
+
+
+@api.route("/aips-by-file-format/<storage_service_id>")
+class AIPsByFormatList(Resource):
+    @api.doc(
+        "list_aips_by_format",
+        params={
+            fields.FIELD_FILE_FORMAT: {
+                "description": "File format name (must be exact match)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_ORIGINAL_FILES: {
+                "description": "Return data for original files (default) or preservation derivatives",
+                "in": "query",
+                "type": "bool",
+            },
+        },
+    )
+    def get(self, storage_service_id):
+        """List AIPs containing given file format"""
+        file_format = request.args.get(fields.FIELD_FILE_FORMAT, "")
+        original_files = parse_bool(request.args.get(fields.FIELD_ORIGINAL_FILES, True))
+        return report_data.aips_by_file_format(
+            storage_service_id=storage_service_id,
+            file_format=file_format,
+            original_files=original_files,
+        )
+
+
+@api.route("/aips-by-puid/<storage_service_id>")
+class AIPsByPUIDList(Resource):
+    @api.doc(
+        "list_aips_by_puid",
+        params={
+            fields.FIELD_PUID: {
+                "description": "PRONOM ID (PUID)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_ORIGINAL_FILES: {
+                "description": "Return data for original files (default) or preservation derivatives",
+                "in": "query",
+                "type": "bool",
+            },
+        },
+    )
+    def get(self, storage_service_id):
+        """List AIPs containing given format version, specified by PUID"""
+        puid = request.args.get(fields.FIELD_PUID, "")
+        original_files = parse_bool(request.args.get(fields.FIELD_ORIGINAL_FILES, True))
+        return report_data.aips_by_puid(
+            storage_service_id=storage_service_id,
+            puid=puid,
+            original_files=original_files,
+        )
+
+
+@api.route("/agents-transfers/<storage_service_id>")
+class AgentData(Resource):
+    @api.doc("agent_info")
+    def get(self, storage_service_id):
+        """List user agents and their transfers"""
+        return report_data.agents_transfers(storage_service_id=storage_service_id)
--- a/AIPscan/Data/data.py
+++ b/AIPscan/Data/data.py
@@ -1,80 +1,10 @@
 # -*- coding: utf-8 -*-
 
-from AIPscan import db
+"""Data endpoints optimized for providing general overviews of AIPs."""
+
+from AIPscan.Data import _get_storage_service, fields
 from AIPscan.helpers import _simplify_datetime
-from AIPscan.models import AIP, Event, File, FileType, StorageService
-
-VALID_FILE_TYPES = set(item.value for item in FileType)
-
-FIELD_AIP_ID = "AIPID"
-FIELD_AIP_NAME = "AIPName"
-FIELD_AIP_SIZE = "AIPSize"
-FIELD_AIP_UUID = "AIPUUID"
-FIELD_AIPS = "AIPs"
-FIELD_ALL_AIPS = "AllAIPs"
-
-FIELD_COUNT = "Count"
-FIELD_CREATED_DATE = "CreatedDate"
-
-FIELD_DERIVATIVE_COUNT = "DerivativeCount"
-FIELD_DERIVATIVE_FORMAT = "DerivativeFormat"
-FIELD_DERIVATIVE_UUID = "DerivativeUUID"
-
-FIELD_EVENT = "Event"
-
-FIELD_FILES = "Files"
-FIELD_FILE_COUNT = "FileCount"
-FIELD_FILE_TYPE = "FileType"
-FIELD_FILENAME = "Filename"
-FIELD_FORMAT = "Format"
-FIELD_FORMATS = "Formats"
-
-FIELD_INGESTS = "Ingests"
-FIELD_INGEST_START_DATE = "IngestStartDate"
-FIELD_INGEST_FINISH_DATE = "IngestFinishDate"
-
-FIELD_NAME = "Name"
-
-FIELD_ORIGINAL_UUID = "OriginalUUID"
-FIELD_ORIGINAL_FORMAT = "OriginalFormat"
-
-FIELD_PUID = "PUID"
-
-FIELD_RELATED_PAIRING = "RelatedPairing"
-
-FIELD_SIZE = "Size"
-FIELD_STORAGE_NAME = "StorageName"
-
-FIELD_TRANSFER_NAME = "TransferName"
-
-FIELD_USER = "User"
-FIELD_UUID = "UUID"
-
-FIELD_VERSION = "Version"
-
-
-def _get_storage_service(storage_service_id):
-    """Return Storage Service with ID or None.
-
-    Unlike elsewhere in our application, here we do not fall back to
-    a different StorageService if the user-supplied ID is invalid to
-    prevent inaccurate information from being returned.
-
-    :param storage_service_id: Storage Service ID
-
-    :returns: StorageService object or None
-    """
-    return StorageService.query.get(storage_service_id)
-
-
-def _get_username(agent_string):
-    """Retrieve username from the standard agent string stored in the
-    database, normally formatted as:
-
-        * username="test", first_name="", last_name=""
-    """
-    USERNAME = "username="
-    return agent_string.split(",", 1)[0].replace(USERNAME, "").replace('"', "")
+from AIPscan.models import AIP, File, FileType
 
 
 def aip_overview(storage_service_id, original_files=True):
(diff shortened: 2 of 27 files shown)


AIPscan/helpers.py
```diff
@@ -10,6 +11,21 @@ def parse_bool(val, default=True):
         return default
 
 
+def parse_date(date_string):
+    """Parse date parameter string into datetime object or None.
+
+    :param date_string: Date string. Valid format is YYYY-MM-DD.
+
+    :returns: datetime.datetime object or None
```
ross-spencer, line 19: I might consider doing the work of `_prepare_datetimes_for_query(...)` here and returning a consistent data type in this helper. For me there are two main clues/reasons: 

1. Mixing return types might cause complexity upstream. Testing for `None` vs. getting to work on a newly minted `datetime`. 
2. I appreciate the amount of thought gone into the date handling across the PR, but the story as I see it through the code, is that we're largely only dealing with date ranges which I believe these helpers are largely setup for handling going forward. I think we're always going to know with a min or a max. So I might suggest a function signature of, something like:

```python
def parse_datetime_bounds(date_string, Upper=False):
   return datetime (or datetime + 1 day if upper==true) if valid,
   else return datetime.min if invalid, or datetime.max if invalid and upper==true
```
I think we might see a cleaner coding style appear by decoding early, rather than doing a little bit now, and a little bit later.

AIPscan/helpers.py
```diff
@@ -10,6 +11,21 @@ def parse_bool(val, default=True):
...
+    """
+    if isinstance(date_string, str):
+        try:
+            return datetime.strptime(date_string, "%Y-%m-%d")
+        except ValueError:
+            pass
+    return None
```
ross-spencer, line 26: I suspect you can get away with something like:
```python
def parse_date(date_string):
    """Parse date parameter string into datetime object or None."""
    try:
        return datetime.strptime(date_string, "%Y-%m-%d")
    except (ValueError, TypeError):
        pass
```
If you wanted to simplify this a little if the function is kept as is.

ross-spencer:
@tw4l thanks for this. Overall I can see this is a high quality PR, although I have to step into it in more detail. 

I feel like AIP created date is exactly the right date to use here as I think this is specifically AIP level reporting. You might be interested in this ticket I mentioned about the archival storage tab too: https://github.com/archivematica/Issues/issues/731 

I have left some initial comments about the date handling which are good additions and I look forward to seeing applied to other existing reports. 

What I am a little bit stuck on before I start to review in more detail, is the similarity in the shape of the data between the two data functions:

1. `curl -X GET "http://127.0.0.1:5000/api/data/aip-overview/1" -H "accept: application/json"`
2. `curl -X GET "http://127.0.0.1:5000/api/data/format-version-count/1" -H "accept: application/json"`

It reminds me of a conversation before, but the similarity feels too hard to ignore, and there should be something done to (my preference - consolidate, rather than duplicate or remove). So I'd just like to ask more about the rationale for 2 over 1. And I think we need to understand what happens to both now, with respect for what each of them offer. My initial instinct is that 2. should probably not have been created without acknowledgement of 1. I could be wrong, the partial duplication might be fine. Maybe this really is a case of repository level reporting, vs. starting to implement functions to support preservation watch and treatment?

tw4l:
Thank you for the feedback @ross-spencer! As always I appreciate your sharp mind and generosity.

> I feel like AIP created date is exactly the right date to use here as I think this is specifically AIP level reporting. You might be interested in this ticket I mentioned about the archival storage tab too: archivematica/Issues#731

Fantastic, and thanks for the helpful context with that issue! Given that, I will plan to add a commit in this PR that modifies the "Format count report" view to also use the AIP created date, to bring the reports into line with each other.

> I have left some initial comments about the date handling which are good additions and I look forward to seeing applied to other existing reports.

Thanks, these comments make a lot of sense! I agree with your insight that we'd be better off decoding early and always returning a `datetime` object, and will make changes accordingly.

> What I am a little bit stuck on before I start to review in more detail, is the similarity in the shape of the data between the two data functions ... It reminds me of a conversation before, but the similarity feels too hard to ignore, and there should be something done to (my preference - consolidate, rather than duplicate or remove). So I'd just like to ask more about the rationale for 2 over 1.

This is fair, and truthfully it's just the result of an oversight on my part. I started with the UI report and worked backwards from there (i.e. deciding what the report looks like, then writing the Data endpoint accordingly, and finally adding the API endpoint to keep good coverage there). I suspect that the similarity and potential to extend what was already there would have been much more apparent to me if I had worked in the reverse order. But you're also right that we've had similar conversations before, so apologies for making us cover this ground again.

Given what you've pointed out, I see a few potential ways forward:

1. Keep both sets of Data endpoints and API endpoints.
2. Remove the new Data and API endpoints, and use the existing `Data.aip_overview` endpoint (adding file size, date limiting, and testing comparable to `Data.format_versions_count`) for the "Format versions count report" in the UI.
3. Hybrid approach: Keep the new `Data.format_versions_count` Data endpoint and continue to use it for the "Format versions count report" in the UI, but remove the new API endpoint in favour of the existing `aip_overview` API endpoint. We could potentially also supplement the `aip_overview` API endpoint by adding aggregate file size, date limiting, and tests so that nothing's lost to consumers by dropping the `format_version_count` API endpoint.

I agree with you that having both API endpoints as they are doesn't add much value currently (your thought about "repository level reporting, vs. starting to implement functions to support preservation watch and treatment" is intriguing, but I'm not sure we're there yet) so that rules out approach 1 for me.

My first thought was to go with approach 2, even though it would mean throwing out a lot of the work in this PR. However, as this would require first fetching data from the `Data.aip_overview` endpoint and then (in another `Data` endpoint function or the `Reporter` view) significantly transforming the shape of that data and sorting it for use in the UI template, there would be costs in terms of speed and complexity vs. leaning on the database/ORM to do the heavy lifting and return the data in the shape we need. Personally, I don't see the benefits of this approach outweighing those costs.

We could also consider the inverse of approach 2 (i.e. keeping the new endpoints and removing the old), but I see value in keeping the API endpoints a bit generic so that other systems can consume them and use them for their own purposes and I think the `aip_overview` API endpoint exemplifies that well.

So as I write this, I'm inclined to go with the hybrid approach 3. In essence, I see this as optimizing for speed in the reporting UI and thoroughness/generality in the API endpoint. What do you think? Does that sound reasonable to you?

ross-spencer:
Thanks @tw4l for laying these options out. It's really helpful. I'm thinking through them. I was thinking it might benefit me to share some previous context/thinking here in person and I wonder impact any particular approach might have one some of the decisions we make in development moving forward, so I requested a Google Hangout with yourself and Peter if that sounds good? 

Going through some of this again I don't feel is that big a deal. It's really timely. There's more that I could have done here to signpost intentions - and what I think might be a somewhat orphaned function in the data module/API endpoint certainly doesn't help! It's going to come up more for other folks as we get others contributing.

tw4l:
Sounds good, thanks @ross-spencer! Looking forward to discussing further this afternoon.

tw4l:
To sum up how we left the offline conversation:

- I will move the `format_version_count` Data and API endpoints to a new module and namespace respectively so that there is a clear differentiation between endpoints that are for generic data consumption vs. endpoints that are feeding raw data for reports implemented in the UI.
- I will assess the other existing endpoints for inclusion in the new module/namespace and move the ones that should be moved - e.g. the largest file report endpoints.
- @ross-spencer will complete code review for this PR, with the understanding that I will be moving the endpoints and making the suggested changes to the helper for parsing date strings into `datetime` objects, but have not done so yet.

AIPscan/Reporter/report_format_versions_count.py
```diff
@@ -0,0 +1,41 @@
...
+    ]
+
+    return render_template(
+        "report_format_versions_count.html",
+        storage_service_id=storage_service_id,
+        storage_service_name=version_data.get(data.FIELD_STORAGE_NAME),
+        columns=translate_headers(headers),
+        versions=versions,
+        total_file_count=sum(version.get(data.FIELD_COUNT, 0) for version in versions),
+        total_size=sum(version.get(data.FIELD_SIZE, 0) for version in versions),
```
ross-spencer, line 37: I'm getting a stack trace error here:
```
Traceback (most recent call last):
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask_restx/api.py", line 639, in error_router
    return original_handler(e)
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/ross-spencer/git/artefactual-labs/aipscanvenv/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/ross-spencer/git/artefactual-labs/AIPscan/AIPscan/Reporter/report_format_versions_count.py", line 37, in report_format_versions_count
    total_size=sum(version.get(data.FIELD_SIZE, 0) for version in versions),
```
It's not coming to me tonight why I might be getting `None` but I am wondering if you have an instinct? We might also need to capture that in tests?

AIPscan/Reporter/templates/report_format_versions_count.html
```diff
@@ -0,0 +1,51 @@
...
+        </tr>
+      </thead>
+      <tbody>
+      {% for version in versions %}
+        <tr>
+          <td>{{ version["PUID"] }}</td>
+          <td>{{ version["Format"] }}</td>
+          <td>{{ version["Version"] }}</td>
+          <td>{{ version["Count"] }}</td>
+          <td>{{ version["Size"] | filesizeformat }}</td>
```
ross-spencer, line 42: This might be related to the reporter module code, but here, I have a `{'PUID': 'fmt/468', 'Format': 'ISO Disk Image File', 'Version': None, 'Count': 1, 'Size': None}` result and the template expects a number to work with and not `None` so I wonder how we can fortify this or the data coming out of the DB (or into) which can help us to avoid this? I looked at the original METS and it looks like the `ISO Disk Image File` was ingested as a standard transfer from one of my FPR test sets - but it's likely the image from the sample data dirs.

AIPscan/Data/data.py
```diff
@@ -224,6 +254,84 @@ def derivative_overview(storage_service_id):
...
+        version_info = {}
+
+        version_info[FIELD_PUID] = version.puid
+        version_info[FIELD_FORMAT] = version.file_format
+        try:
+            version_info[FIELD_VERSION] = version.format_version
+        except AttributeError:
+            pass
+        version_info[FIELD_COUNT] = version.file_count
+        version_info[FIELD_SIZE] = version.total_size
```
ross-spencer, line 328: Ah, it looks like maybe we can test for none here? (see template comment)

AIPscan/Data/data.py
```diff
@@ -224,6 +254,84 @@ def derivative_overview(storage_service_id):
...
 
 
+def _format_versions_count_query(storage_service_id, start_date=None, end_date=None):
+    """Fetch information from database on format versions.
+
+    :param storage_service_id: Storage Service ID.
+    :start_date: Optional inclusive AIP creation start date
+    (datetime.datetime object).
+    :end_date: Optional inclusive AIP creation end date
+    (datetime.datetime object).
```
ross-spencer, line 264:
```suggestion
    :param storage_service_id: Storage Service ID.
    :start_date: Optional inclusive AIP creation start date
        (datetime.datetime object).
    :end_date: Optional inclusive AIP creation end date
        (datetime.datetime object).
```
I am still getting used to this more complex style of docstring. Admittedly I haven't spent a lot of time looking at this, but what do you think about adding a hanging indent: https://stackoverflow.com/a/50455731 - to help readability? It would be good to do that wherever we break a docstring over two lines. Initially it's hard to pick up as part of the same string. 

Is there another idiomatic approach that you happen to know about?

ross-spencer (commented):
@tw4l lots of this looking good. I'm submitting a first go code-review in the hope it might aid any of your efforts this week. I know you're working on the changes we discussed. There are a few small style things here, something to think about in the tests, and possibly some fortification to add to the report generation but I like the look of the report.

AIPscan/Data/data.py
tw4l: Yes I think you're right!

AIPscan/Data/data.py
tw4l: These detailed docstrings are a new-ish habit for me so I don't have a ready answer for other approaches but this is a good suggestion and I agree it would help a lot with readability.

AIPscan/Reporter/templates/report_format_versions_count.html
tw4l: I think it's likely that this `File` object has the format and size it does because of this potentially misleading `AttributeError` handling: https://github.com/artefactual-labs/AIPscan/blob/39dbf8801e0b52e58f6e3f1cc9f438929849b85e/AIPscan/Aggregator/database_helpers.py#L196-L200 (see also issue #83). I wonder if it's actually an ISO disk image...

Regardless, I think you hit the nail on the head that we should be checking for `None` in the data endpoint to avoid this issue. Good spot!

AIPscan/Reporter/report_format_versions_count.py
tw4l: I will add the check for `None` in the Data endpoint here, but agree that it would be good to capture this in tests! I will work on including that in my revisions.

AIPscan/Reporter/templates/reports.html
```diff
@@ -54,8 +51,8 @@
         <tr>
             <td>Format version count</td>
             <td></td>
-            <td></td>
-            <td></td>
+            <td><input type="text" id="startdate3" value="{{ start_date }}"></td>
```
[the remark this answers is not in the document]
tw4l: Yes, good point!

AIPscan/Data/tests/test_format_versions_count.py
```diff
@@ -0,0 +1,165 @@
...
+
+TOTAL_FILE_SIZE = JPEG_1_01_FILE_SIZE + JPEG_1_02_FILE_SIZE
+
+DATE_BEFORE_AIP_1 = "2019-01-01"
+DATE_AFTER_AIP_1 = "2020-01-02"
+DATE_BEFORE_AIP_2 = "2020-05-30"
+DATE_AFTER_AIP_2 = "2020-06-02"
+
+
+class MockFormatVersionsCountQueryResult:
```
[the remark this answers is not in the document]
tw4l: Oh interesting, this is new to me. Thanks for sharing!

AIPscan/Data/tests/conftest.py
```diff
@@ -94,3 +105,93 @@ def app_with_populated_files(scope="package"):
...
         db.drop_all()
+
+
+@pytest.fixture
+def app_with_populated_format_versions(scope="package"):
+    """Fixture with pre-populated data.
+
+    This fixture is used to create expected state which is then used to
+    test the Data.format_versions_count endpoint.
+    """
```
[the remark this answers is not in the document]
tw4l: Oh I like this a lot @ross-spencer. Thanks for putting in a lot of effort and good thinking here! I'll do my best to get this into my revisions but appreciate the option to kick it to a future commit/PR if needed.

AIPscan/Reporter/templates/report_format_versions_count.html
tw4l: Ah whoops, just re-read your comment and my "I wonder if it's actually an ISO disk image..." was off base here. Sorry about that! :sweat_smile:

AIPscan/Data/tests/conftest.py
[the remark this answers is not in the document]
tw4l: Helpers added in commit dbe8350.

AIPscan/Reporter/templates/reports.html
```diff
@@ -227,6 +204,28 @@
...
+                }
+            }
+        });
+        $("#enddate3").datepicker({
+            dateFormat: 'yy-mm-dd',
+            onSelect: function(date) {
+                var startdate = $('#startdate3').val();
+                if (startdate != "") {
+                    if (startdate > date) {
+                        alert("The start date must be before the end date");
```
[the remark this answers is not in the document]
tw4l: I've added two new consts here - `DATE_ALERT_START` and `DATE_ALERT_END` - for the two variations of the message

tw4l:
Hi @ross-spencer - I believe I've now responded to each of your suggestions and comments, and that this PR is now ready for code review. Thanks again for the great suggestions here!

In addition to responding to the in-line comments, I've moved the report endpoints to a new `Data.report_data` module and `API.namespace_report_data` namespace. This required a lot of changes in the `Data`, `API`, and `Reporter` blueprints and updates to the tests, so the diff might be a bit larger. I've made sure that all tests are passing and done a quick test of each of the API endpoints and UI reports to make sure everything looks good following the change.

Along the way I discovered a bug in the `agents_transfers` endpoint that sometimes cause that report and API endpoint to raise an `AttributeError`, which I filed an issue for here: https://github.com/artefactual-labs/AIPscan/issues/104.

tw4l:
I realized right after marking this PR as ready for re-review that I forgot to update the "File format count" report to use `AIP.create_date` for date filtering, so I've added a follow-up commit with that change.

ross-spencer (approved):
One question around style/presentation for the API URL, but approving, and: 

santa

The way you approached moving things about made it really simple to re-review. Lots of this looking really clean and easy to use now too. Thanks @tw4l

AIPscan/API/namespace_report_data.py
```diff
@@ -0,0 +1,142 @@
...
+
+from flask import request
+from flask_restx import Namespace, Resource
+
+from AIPscan.helpers import parse_bool, parse_datetime_bound
+from AIPscan.API import fields
+from AIPscan.Data import report_data
+
+api = Namespace(
+    "report_data", description="Retrieve data optimized for AIPscan reports"
```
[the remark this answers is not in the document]
tw4l: Ah, good catch! I agree the dash is nicer and more consistent. I'll make that change before merging.

PR #99, opened 2020-12-03
Revert incorrect handling of FAILURE status (fetch job statuses)

### LINKED ISSUE #72: Make Fetch Job activity display more meaningful
In an earlier version of AiPscan the Fetch Job output console posted a notice of how many METS flles it was going to download and then numbered each METS file beingn downloaded so the user could get a sense of how far in the Fetch Job AIPscan is. This seems to have dissapeared in more recent commits:

Screen Shot 2020-10-07 at 2 38 46 PM

A similar numbering scheme was used to show progress in Parsing METS files but now multiple numbers are shown per Parsing job and are restarted for each new parsing job. This gives the user no sense of actual progress. 

The only thing the job console is useful for now is indicating that the server is actually busy and hasn't hung. Either we fix the display to re-introduce numbering or something similar or we should simplify the console to just have a "processing" indicator that is either green or red to indicate activity or a failure.

### LINKED ISSUE #73: Multiple Flash messages appearing for a single successful Fetch Job
There should be only one per Fetch Job:

Screen Shot 2020-10-07 at 2 49 28 PM

ross-spencer:
FAILURE handling was added incorrectly to package_list_task_status.
This commit reverts those changes to make the console output work
as it did previously. This also quietens down the completed tasks
alerts as well.

Resolves #72 
Resolves #73 

@peterVG first, apologies for my speaking out of term the other day. I had diagnosed this wrong. Looking again tonight I have discovered a mis-reading of the original JavaScript functions you had created when I changed them here: https://github.com/artefactual-labs/AIPscan/commit/e59056e9ffa01213fa6517da898946493e13c5e3#diff-606a3136c88e560ab867ae3abac4c67013a7a577f7b69878f20c7939fa99e179. I have not tried to add FAILURE handling in again as this might be something you do differently with your current work, and so this just reverts the change back to a working JavaScript. If you have an opportunity to test and review we'll get this merged and it should be one less annoyance :slightly_smiling_face: 

_Recorded tonight:_

Peek 2020-12-02 23-07

### DIFF
--- a/AIPscan/static/js/tasks.js
+++ b/AIPscan/static/js/tasks.js
@@ -17,29 +17,20 @@ function new_fetch_job(storageServiceId){
 }
 
 function package_list_task_status(taskId, showcount, fetchJobId){
-  const inspect_logs = ": Please inspect the celery logs"
   const status_pending = 'PENDING'
   const status_progress = 'IN PROGRESS'
-  const status_failure = "FAILURE"
   $.ajax({
     type: 'GET',
     url: '/aggregator/package_list_task_status/' + taskId,
     datatype: "json",
     success: function(data) {
-      state = data['state']
-      if (state != status_pending && state != status_progress && state != status_failure) {
+      let state = data['state'];
+      if (state != status_pending && state != status_progress) {
         $('#console').append('<div class="log">' + state +'</div>')
         $('#console').append('<div class="log">Downloading AIP METS files</div>')
         get_mets_task_status(data["coordinatorId"], 0, fetchJobId);
       }
-      if (state == status_failure) {
-        err = "".concat('<div class="log">', state, inspect_logs ,'</div>')
-        $('#console').append(err)
-        return
-      }
       else {
-        output = "".concat('<div class="log">', state, ': Downloading AIP METS files</div>')
-        $('#console').append(output)
         if (showcount == false) {
           if ('message' in data){
             $('#console').append('<div class="log">' + data["message"] +'</div>')


AIPscan/static/js/tasks.js
```diff
@@ -17,29 +17,20 @@ function new_fetch_job(storageServiceId){
...
   const status_progress = 'IN PROGRESS'
-  const status_failure = "FAILURE"
   $.ajax({
     type: 'GET',
     url: '/aggregator/package_list_task_status/' + taskId,
     datatype: "json",
     success: function(data) {
-      state = data['state']
-      if (state != status_pending && state != status_progress && state != status_failure) {
+      state = data['state'];
```
tw4l, line 27: You may want to initialize `state` with `let state = ...`. Otherwise I think loose mode JS will make this a global variable and strict mode will throw an error.

ross-spencer: Ah, yep. Language switching failings.

tw4l: Yeah no worries, that happens!

tw4l (approved):
LGTM! Code changes make sense and I'm seeing the expected UI behaviour on my end.

image

`tasks.py` could benefit from some linting (I see missing semi-colons in places not touched by your changes in these commits, for instance) but I think touching that up might be out of scope for this PR, so I'm approving as-is.

ross-spencer:
Thanks @tw4l! Glad to have this fixed. I've logged two extra issues around JavaScript and also related to your suggestions for `fontawesome` as well. Primary one for JS is: https://github.com/artefactual-labs/AIPscan/issues/103 if you have any favorite linters/experience then it would be great to get your feedback there too. I just quickly tried on or two tonight - `eslint` didn't like being run on my machine! JSlint seems older? But it got the job done as a demonstration. Anyhoo!

Issue #31, opened 2020-07-08
Let's use isort


ross-spencer:
We have a proposed ADR: https://adr.archivematica.org/0004-isort-import-ordering.html in general I do think it makes life easier, and it's one less thing to make CR subjective. And one less thing to introduce inconsistency across modules. 

I ran `pylint` before and after, and the score was nice to read:

* Cmd: `artefactual-labs/AIPscan/AIPscan$ isort $(find * | grep .py$)`
* Pylint output: `Your code has been rated at 7.32/10 (previous run: 6.99/10, +0.33)`

peterVG:
Okay, sounds good. Should we run it as one repo-wide batch and merge? Too bad we don't have tests to run afterwards. I'm a little concerned about circular imports (e.g. an import being called too early/late). I ran into these when re-factoring to blueprints. But that was basically confined to this file: https://github.com/artefactual-labs/AIPscan/blob/main/AIPscan/__init__.py So I know where to look if things break after the isort.

ross-spencer:
>  Should we run it as one repo-wide batch and merge? 

Perfect!

**Minus `__init__.py`**

Hehe! I can can confirm that circular imports in that file will cause a problem, but also, I think that just means initially, we can leave that particular `__init__.py` out of the `isort` mix. 

I feel it would be good to get the circular imports resolved in time too. I tend to look at these as clues for improving architecture. I did have a look last night before I saw your update. I'm not familiar enough with the code yet to know the objects and the way they're being passed around to have a solution for this. I started looking at [this blog](https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/) which gives an SQLAlchemy example but I didn't quite complete that trail of thought before moving onto another piece. It looks like maybe the celery context manager is the only other one impacted. I'll read up a little more on the Blueprint and factory pattern.

![warhold-factory](https://media3.giphy.com/media/fxOk71qCsFIqLpBTSu/giphy.gif?cid=ecf05e4779908f552070dfe8f063a3b019051418d0029eab&rid=giphy.gif)

ross-spencer:
With linting errors now resolved I had a brief look into this. It seems like the main concern is making isort compatible with Black, however, with [these options](https://black.readthedocs.io/en/stable/compatible_configs.html#isort) the isort defaults may be good enough (should be, i.e. it is less effort to maintain and already de-facto standardised). 

[...]

PR #105, opened 2020-12-30
isort to improve automation of code-review and code-consistency

ross-spencer:
Connected to #31

### DIFF
--- a/AIPscan/Aggregator/views.py
+++ b/AIPscan/Aggregator/views.py
@@ -1,24 +1,20 @@
 # -*- coding: utf-8 -*-
 
-from datetime import datetime
 import os
 import shutil
+from datetime import datetime
 
 from celery.result import AsyncResult
-from flask import Blueprint, render_template, redirect, request, flash, url_for, jsonify
+from flask import Blueprint, flash, jsonify, redirect, render_template, request, url_for
 
 from AIPscan import db
-from AIPscan.Aggregator.task_helpers import get_packages_directory
-from AIPscan.Aggregator.forms import StorageServiceForm
 from AIPscan.Aggregator import tasks
+from AIPscan.Aggregator.forms import StorageServiceForm
+from AIPscan.Aggregator.task_helpers import get_packages_directory
 from AIPscan.extensions import celery
-from AIPscan.models import (
-    FetchJob,
-    StorageService,
-    # Custom celery Models.
-    package_tasks,
-    get_mets_tasks,
-)
+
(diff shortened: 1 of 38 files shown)


AIPscan/Aggregator/views.py
```diff
@@ -1,23 +1,22 @@
...
 
 from AIPscan import db
-from AIPscan.Aggregator.task_helpers import get_packages_directory
-from AIPscan.Aggregator.forms import StorageServiceForm
 from AIPscan.Aggregator import tasks
+from AIPscan.Aggregator.forms import StorageServiceForm
+from AIPscan.Aggregator.task_helpers import get_packages_directory
 from AIPscan.extensions import celery
-from AIPscan.models import (
+from AIPscan.models import (  # Custom celery Models.
```
tw4l, line 15: This comment ended up in a funny place. Perhaps it could be removed?

tox.ini
```diff
@@ -23,3 +23,11 @@ ignore =
...
 
 import-order-style = pep8
+
+[isort]
+multi_line_output = 3
+include_trailing_comma = True
+force_grid_wrap = 0
+use_parentheses = True
+ensure_newline_before_comments = True
+line_length = 88
```
tw4l, line 33: These are to get `isort` to work nicely with `black`, yeah? Nice attention to detail!

tw4l (approved):
Nice work, @ross-spencer! I left one comment where applying `isort` put a comment in an funny spot - otherwise, this looks great. It's a really nice addition to the project, and the approach of tying into a pre-commit hook is :man_cook: :ok_hand:

That only leaves one thing to say...

LGTM image

AIPscan/Aggregator/views.py
ross-spencer: Oh weird, that may have started in a weird place too? No idea why! Correcting it seems to have put it back in a sensible place. Great catch!

**Before**

```python
from AIPscan.models import (	
    FetchJob,	# Custom celery Models.
    StorageService,	from AIPscan.models import FetchJob, StorageService, get_mets_tasks, package_tasks
    # Custom celery Models.	
    package_tasks,	
    get_mets_tasks,	
)
```

**After**

```python
# Custom celery Models.
from AIPscan.models import FetchJob, StorageService, get_mets_tasks, package_tasks
```

ross-spencer:
Awesome @tw4l! Really glad to have this in here now. Thank you for the review/+1 on this.

PR #107, opened 2021-01-27
Changes favicon color to match AIPScan color scheme

scollazo:
This also helps finding the AIPscan in the tab bar...


ross-spencer (approved):
:green_circle: LGTM.

Issue #112, opened 2021-02-20
Test: valid Storage Service URL, username, and API key are used


peterVG:
Related to Issue https://github.com/artefactual-labs/AIPscan/issues/111

Problem: can't expose a live SS URL & API Key in this project's Git repo. 
-- Should we use ENV variables? Is that too awkward to support across deployments?
-- Could we prepare stubs/fixtures for this test instead? Doesn't it need to make a live `requests` call to be a valid test?

tw4l:
Resolved in https://github.com/artefactual-labs/AIPscan/commit/c4eb77756d43897d9c1b9cf58acd19c6898dd9b2

PR #113, opened 2021-02-21
add Storage Service connection test

peterVG:
Addresses Issue https://github.com/artefactual-labs/AIPscan/issues/112 

NOTE: used local ENV variables to get around issue of storing authentication secrets in Git repo. Discuss?

Successfully failed and passed this test when changing the URL, username, API Key to Null or invalid values and vice versa.

### DIFF
--- a/AIPscan/Aggregator/tests/test_storage_service.py
+++ b/AIPscan/Aggregator/tests/test_storage_service.py
@@ -1,0 +1,30 @@
+import os
+
+from AIPscan.Aggregator import tasks, task_helpers
+
+# Storage service URLs, usernames and API keys are sensitive information.
+# To run this test, create yours as local ENV variables using the following
+# names: 'AIPSCAN_SS_URL', 'AIPSCAN_SS_USER', and 'AIPSCAN_SS_KEY'
+SS_URL = os.getenv('AIPSCAN_SS_URL')
+SS_USER = os.getenv('AIPSCAN_SS_USER')
+SS_KEY = os.getenv('AIPSCAN_SS_KEY')
+
+
+def test_ss_connection():
+    """Test that a connection to the Storage Service server can be made
+    successfully and that there are no incorrect values for the Storage Service URL, username or API Key"""
+
+    if SS_URL is None or SS_USER is None or SS_KEY is None:
+        assert False, "Missing Storage Service URL, username, or API Key in ENV variables."
+    else:
+        api_url = {
+            "baseUrl": SS_URL,
+            "userName": SS_USER,
+            "apiKey": SS_KEY,
+            "offset": 0,
+            "limit": 20,
+        }
+
+        base_url, request_url_without_api_key, request_url = task_helpers.format_api_url_with_limit_offset(api_url)
+
+        assert tasks._make_request(request_url, request_url_without_api_key)


ross-spencer:
This is starting to look like a good test @peterVG. Striking the balance between something more like an integration test and a unit test is quite tough sometimes - finding the boundaries between realistic testing and mocking too. 

It sounds like you're aiming for functionality along the lines of `vcrpy`: https://github.com/kevin1024/vcrpy - our easiest code using this is amclient: https://github.com/artefactual-labs/amclient. Put simply vcrpy records the http request and response between the client and the server and enables you to play those back during a unit test. It mocks a live request so it gives you that realistic outcome that you're looking for. Once parameterized you can ask the test run your two or more scenarios too to test success and failure automatically.

E.g. example cassette recordings:

* [request (locations in storage service)](https://github.com/artefactual-labs/amclient/blob/64111b6195c18f4f3721a9de75dcb2c5dc2487f9/fixtures/vcr_cassettes/create_location.yaml#L2-L20)
* [and response (lists locations)](https://github.com/artefactual-labs/amclient/blob/64111b6195c18f4f3721a9de75dcb2c5dc2487f9/fixtures/vcr_cassettes/create_location.yaml#L21-L50)

NB. vcrpy does have its drawbacks though (my opinion) and they don't really protect against a shift in code on he back-end. 
https://github.com/archivematica/Issues/issues/530

From a unit test perspective you can achieve something similar too but it might not feel as realistic. What you want to do with the unit test is check that the response from AIPscan is something sensible. So you would maybe mock a connection and ask it to return true or false or raise an exception. For either you might just want to make sure that the success or failure outcome of the function is anticipated e.g. testing "connection successful" returns or "connection failure" (pseudo code). 

I feel this work can be shaped into either pretty easily. And happy to chat more in the week about that with y'all. 

One additional note is that we don't get a lot of feedback about failures on GitHub so you just need to run  `tox -e linting` on this to correct the errors. (You may get an email I don't know or may need to register an email to get failure reports sent to).

tw4l:
Hey @peterVG! I don't want to repeat too much what @ross-spencer has said above and I don't have much experience with `vcrpy`, but I will say that guiding these tests toward unit tests can make their implementation a lot simpler.

One thing to consider re: the behaviour the end user is seeing here that we want to fix and add tests for - it seems like  `_make_request` is doing what we'd expect it to do and raising a `TaskError` when improper Storage Service credentials are provided, but our handling of that exception is working quite as expected as desired (i.e. not stopping the Celery task appropriately). That's certainly something we could sort out in the Celery tasks, but I'd also encourage you to check the Storage Service credentials prior to kicking off the Celery tasks.

For example, we could use `_make_request` to make a test GET request to e.g. `/api/v2/file/` right [at the beginning of the `Aggregator.views.new_fetch_job` view](https://github.com/artefactual-labs/AIPscan/blob/6f8f18acfa8db03f464786e965c4b44160f85f25/AIPscan/Aggregator/views.py#L142-L154), just after defining the `api_url` dictionary. And then if it raises a `TaskError`, log an error, quit, and provide UI feedback accordingly before kicking off a Celery task. 

Something like (just a rough idea):
```python
def test_storage_service_connection(api_url):
    """Test Storage Service credentials.
    
    :param api_url: Storage Service credentials (dict)

    :raises ConnectionError: if credentials are invalid
    """
    # Make a new dict instead of altering our actual credentials.
    test_credentials = dict(api_url)
    # Set limit low to help prevent blocking for unnecessarily long.
    test_credentials["limit"] = 10
    _, _, test_url = task_helpers.format_api_url_with_limit_offset(test_credentials)
    try:
        _ = tasks._make_request(test_url)
    except TaskError as err:
        raise ConnectionError(str(err))


@aggregator.route("/new_fetch_job/<id>", methods=["POST"])
def new_fetch_job(id):
    storage_service = StorageService.query.get(id)
    api_url = {
        "baseUrl": storage_service.url,
        "userName": storage_service.user_name,
        "apiKey": storage_service.api_key,
        "offset": str(storage_service.download_offset),
        "limit": str(storage_service.download_limit),
    }
    try:
        test_storage_service_conection(api_url)
    except ConnectionError as err:
        error_msg = str(err)
        logger.error(error_msg)
        response = {"error": error_msg}
        return jsonify(response), 400  # Return 400 Bad Request
    ...    
```

Then potentially this failure scenario could be covered by a few unit tests, to make sure that e.g. `tasks._make_request` raises a `TaskError` if the response code from the Storage Service is anything other than 200 (you could mock the response from the Storage Service to isolate just the handling behaviour of `_make_request`) and that the view returns 400 early on a `ConnectionError`.

tw4l:
Closed in favour of https://github.com/artefactual-labs/AIPscan/pull/128

PR #114, opened 2021-02-21
display Fetch Job stats via icon click

peterVG:
Making use of the Fetch Job statistics that are already being stored in the database, this commit displays them on the Storage Service template. Rather than clutter the screen with stats for multiple Fetch Jobs, it only reveals the stats when clicking an Info icon. Clicking the icon again re-hides the stats info.

Connected to https://github.com/artefactual-labs/AIPscan/issues/47


AIPscan/Aggregator/templates/storage_service.html
```diff
@@ -49,7 +49,16 @@
       <tr>
         <td>{{ mets_fetch_job.download_start }}</td>
         <td>{% if mets_fetch_job.download_end %} {{ mets_fetch_job.download_end - mets_fetch_job.download_start }} {% endif %}</td>
-        <td>{{ mets_fetch_job.total_packages }}</td>
+        <td>{{ mets_fetch_job.total_packages }} packages
+          <i class="fa fa-info-circle fa-lg" id="{{ mets_fetch_job.id }}" style="margin-left: 5px;"></i><br/>
+          <div id="stats-{{ mets_fetch_job.id }}" class="stats" style="font-size:80%; color: #999;">
```
tw4l, line 54: I'd recommend using Bootstrap tags and classes instead of a custom `style` tag to manage the styles. So for instance,  inside this `<div>` you could do something like
```
<span class="text-muted">
  <small>
    {{ mets_fetch_job.total_aips }} AIPs<br>
    {{ mets_fetch_job.total_sips }} SIPs<br>
    {{ mets_fetch_job.total_dips }} DIPs<br>
    {{ mets_fetch_job.total_replicas }} replicated AIPs<br>
    {{ mets_fetch_job.total_replicas }} deleted AIPs
  </small>
</span>
```
We do that elsewhere in the code base as well, e.g.: https://github.com/artefactual-labs/AIPscan/blob/main/AIPscan/Reporter/templates/report_largest_files.html#L74. The main advantage this change would give is if we want to change what our `text-muted` style looks like, we do so in one place and have it apply to the whole project (and full transparency, on my mind because I think the low contrast might be a bit of an accessibility issue, and so we might want to find a less-muted shade of grey).

AIPscan/Aggregator/templates/storage_service.html
```diff
@@ -49,7 +49,16 @@
       <tr>
         <td>{{ mets_fetch_job.download_start }}</td>
         <td>{% if mets_fetch_job.download_end %} {{ mets_fetch_job.download_end - mets_fetch_job.download_start }} {% endif %}</td>
-        <td>{{ mets_fetch_job.total_packages }}</td>
+        <td>{{ mets_fetch_job.total_packages }} packages
+          <i class="fa fa-info-circle fa-lg" id="{{ mets_fetch_job.id }}" style="margin-left: 5px;"></i><br/>
```
tw4l, line 53: For HTML5, the preferred line break tag is `<br>`, no trailing space/slash needed: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

AIPscan/Aggregator/templates/storage_service.html
```diff
@@ -49,7 +49,16 @@
...
+          <div id="stats-{{ mets_fetch_job.id }}" class="stats" style="font-size:80%; color: #999;">
+            {{ mets_fetch_job.total_aips }} AIPs <br />
+            {{ mets_fetch_job.total_sips }} SIPs <br />
+            {{ mets_fetch_job.total_dips }} DIPs <br />
+            {{ mets_fetch_job.total_replicas }} replicated AIPs <br />
```
tw4l, line 58: Nit: We often call these "replicated AIPs", but it's a bit ambiguous whether that means the original or the replica. I think it might be more accurate or at least unambiguous to call them "AIP replicas".

AIPscan/Aggregator/templates/storage_service.html
```diff
@@ -67,6 +76,13 @@
   $('#newfetchjob').on('click', function () {
     new_fetch_job({{ storage_service.id }});
     });
+  $('.stats').hide();
+  $('.fa-info-circle').click(function() {
+    if($('#stats-'+this.id).is(':visible'))
+        $('#stats-'+this.id).hide();
+    else
+        $('#stats-'+this.id).show();    
+    });
```
tw4l, line 85: Succinct, readable, and effective - nice job!

One minor nit/linting issue is that it'd be great to have opening and closing curly braces aligned - these are fairly small closures so not an immediate practical need maybe, but having them properly aligned really helps with readability when code grows. In practical terms I think it would just mean moving the closing `});`s here two spaces to the left.

tw4l: This is the sort of thing that we will ideally eventually have a JS linter to help spot (and potentially even auto-correct for us, like `black` does for our Python). And it's absolutely a nit, so I'll leave it up to you to decide if it's worth the time.

tw4l (approved):
Hi @peterVG - this is looking great! I've left some suggestions and a few nits, but this is already solid so I'm approving and you can make those changes before squashing and merging. Nice work! :muscle:

AIPscan/Aggregator/templates/storage_service.html
peterVG: Yes, custom inline CSS is a bad temp hack. Thanks for this cleanup suggestion. Done.

AIPscan/Aggregator/templates/storage_service.html
peterVG: themoreyouknow

peterVG: I learned basic HTML back in the 90's and this was considered very important (yet totally irrelevant in the end) practice.

AIPscan/Aggregator/templates/storage_service.html
peterVG: +1 Yes agreed.

peterVG: Also, great nit-picking. I do this kind of semantic hair splitting all the time and appreciate it when others do it too. Accuracy is in our field is important. Starting with how we describe the tools that run our systems.

AIPscan/Aggregator/templates/storage_service.html
peterVG: I'll take any suggestions on improving the readability of Javascript / JQuery, lol. I've moved the bracket as suggested. Have we shortlisted JS Linters yet? I'll gladly install one.

Issue #104, opened 2020-12-17
Problem: The agent_transfers Data endpoint throws an AttributeError for AIPs with files extracted from a package


tw4l:
While refactoring the Data modules for https://github.com/artefactual-labs/AIPscan/pull/98, I discovered that the `agent_transfers` Data endpoint throws an `AttributeError` if an AIP consists exclusively of files extracted from a package that is itself deleted after extraction. At first I thought this was related to my changes, but I'm seeing the same behaviour in the `main` branch.

Traceback:

```
Traceback (most recent call last):
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask_restx/api.py", line 638, in error_router
    return original_handler(f)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask_restx/api.py", line 636, in error_router
    return self.handle_error(e)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask_restx/api.py", line 638, in error_router
    return original_handler(f)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/twalsh/dev/AIPscan/venv/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
[...]

PR #122, opened 2021-04-06
Add defensive check for ingestion events to prevent 500 errors with Agent-Transfers report

tw4l:
This is a temporary solution to prevent `AttributeErrors` on `None` ingestion events (a current result of packages deleted after extraction in Archivematica) from preventing display of the Agent-Transfers report, until a better representation of deleted packages is possible in the AIPscan database.

Connected to #104

### DIFF
--- a/AIPscan/Data/report_data.py
+++ b/AIPscan/Data/report_data.py
@@ -330,6 +330,10 @@ def agents_transfers(storage_service_id):
             .filter(File.aip_id == aip.id, Event.type == EVENT_TYPE)
             .first()
         )
+        # This defensive check is necessary for now because of packages that
+        # are deleted after extraction. See issue #104 for details.
+        if event is None:
+            continue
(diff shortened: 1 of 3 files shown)


peterVG:
The tests pass. Nice work including those, thanks. The bug where Agent Transfer reports are not appearing is fixed. 

The tabular report works well. However, there looks to be a (unrelated) bug for the Gant chart report. It is not displaying any data.

Screen Shot 2021-04-06 at 1 25 13 PM

We can probably file that separately though and accept this commit first.

peterVG (approved):

tw4l:
Thanks @peterVG! We discussed offline - it appears that the transfers are being charted in the Gantt chart but their small size size to the overall timeline (e.g. a 30-minute ingest on a 1 year timeline) means they're sometimes not visible. That is best represented as a separate issue.

Issue #121, opened 2021-04-06
Problem: "File format count" report's header information isn't date-limited


tw4l:
**Expected behaviour**

Report header metadata is accurate to the parameters specified for the report.

**Actual behaviour**

The amount of files in the header is the total amount, not limited to the requested time period.

PR #123, opened 2021-04-06
Fix file count in "File format count" header

tw4l:
Ensure that original file count in the header is limited by the date parameters specified for the report.

Connected to #121

### DIFF
--- a/AIPscan/Reporter/report_formats_count.py
+++ b/AIPscan/Reporter/report_formats_count.py
@@ -36,7 +36,6 @@ def report_formats_count():
     originals_count = 0
 
     for aip in aips:
-        originals_count += aip.original_file_count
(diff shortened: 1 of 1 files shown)


peterVG (commented):
Clumsy original error. Thanks for rectifying. Works as advertised during QA. 

The commit warnings look to be related to missing tests for the new lines which are not expected so merging to main LGTM.

peterVG (approved):

tw4l:
Thanks @peterVG! Yeah, to date we haven't been writing tests for the `Reporter` views. This view is a bit exceptional in that it's constructing its own data rather than relying on a `Data` endpoint, so if you want me to add tests here let me know. Otherwise I think I just need your approval to merge.

PR #125, opened 2021-04-08
Prevent duplicate AIP imports

tw4l:
Previously, each new Fetch Job would result in duplicate AIP, file, and event records in AIPscan. By calculating and storing a SHA-256 hash for METS files, this commit removes this duplication. Existing AIP records are left as-is on subsequent Fetch Jobs if the hash of a downloaded METS match a stored value. If an AIP is changed, the old records are deleted from the database prior to creating new AIP and file objects.

The overhead of hashing the METS files doesn't appear to be too significant in my testing, and it means that subsequent fetch jobs after the first are much faster than before since they are only processing new METS files.

Connected to #6


ablwr (approved):
Nice tests and clean-up!!! I trust the tests.

Issue #86, opened 2020-10-28
Download reports as CSV files


tw4l:
At the October 22 sprint meeting, it was requested that reports should be downloadable, preferably as a CSV file

PR #126, opened 2021-04-15
Add CSV export functionality for reports

tw4l:
Connected to #86 

This PR adds CSV export functionality to each of AIPscan's reports which is accessible through a "Download CSV" button in the header of each report (set in a template partial along with the print button, which is implemented across all the reports).

The CSVs are delivered via the `Reporter` views rather than via an API call as I originally suggested. I'm still of two minds about whether this is the right approach but I ended up going this way for a few reasons, namely:

- To more easily match the state of what the user sees in their browser with what is delivered in the CSV
- Because some of the older reports require post-processing in the view in a way that is an impediment to implementing CSV download via the API
- Because the information in most of the CSVs would be duplicative of what's already available in a more flexible format via the existing JSON endpoints
- To provide a hook for future translations of column headers

I've introduced some new helpers to use moving forward and made a number of changes, many of which were necessary to implement CSV export in a way that reflected the browser reports. Such changes include:

- Replacing our existing human-readable file size helper with jinja2's `filesizeformat` (which is also in use in our templates), to bring consistency in our formatted sizes between the browser-rendered reports, printed reports, and CSV exports
- Refactoring the business logic/backend the User Ingest Log and Preservation Derivatives reports to add new `Data.report_data` endpoints and tests for each
- Renaming and reordering some reports in the UI for consistency
- Replacing the `request_params` dict with constants in a new `request_params.py` module

The biggest change is to the frontend of the Preservation Derivatives report, which before changes looked like the following:

pres_derivatives_old

Because there was a table for each AIP on the page, it was difficult to create usable CSV and print reports. I've rewritten this to be in a single table, which by default displays all preservation derivatives in the selected Storage Service but can be filtered by individual AIP.

All AIPs (default):

pres_derivatives

Filtered by AIP:

pres_derivatives_filtered

This is a significant change but one that I think will ultimately make it easier for folks to use the report in its different formats. The fields that have muted subtext (UUIDs/PUIDs) in the browser UI are separated into distinct fields for the CSV:

pres_derivatives_csv


ablwr:
I agree with your approach. But I am a bit opinionated in that I like when the client/browser does as much work as possible, and when we rely on tried-and-true methods rather than writing our own from scratch.

Now I'll dig into the codebase and see if I have more granular suggestions!

AIPscan/API/namespace_data.py
```diff
@@ -25,42 +25,42 @@
...
                 "type": "bool",
             }
         },
     )
     def get(self, storage_service_id):
-        """List PUIDs and AIPs containing examples of them"""
+        """Return data on AIPs and the file formats they contain."""
         original_files = parse_bool(request.args.get(fields.FIELD_ORIGINAL_FILES, True))
-        aip_data = data.aip_overview(
+        aip_data = data.aip_overview_two(
```
ablwr, line 42: This predates your specific work, but can we rename "aip_overview_two" to be more descriptive, rather than just having two similarly-named `aip_overview` and `aip_overview_two` fields? Like `aip_file_formats_overview` or something that gives more context.

AIPscan/Reporter/helpers.py
```diff
@@ -30,25 +36,94 @@ def translate_headers(headers):
...
+
+    :param headers: Row headers (list of str)
+    :param rows: Data to write to CSV, returned from Data endpoint (list of dicts)
+    :param filename: CSV filename (str)
+    """
+    string_io = StringIO()
+    writer = csv.writer(string_io)
+    writer.writerow(headers)
+    for row in rows:
+        _remove_primary_keys(row)
```
ablwr, line 110: Not opposed, just curious: Is there any reason a user might want to know the AIPScan primary keys? Could there be a world in which an exported CSV for be imported again later, like as a migration without having to re-ingest every AIP? And would the primary keys even matter?

AIPscan/Reporter/report_aip_contents.py
```diff
@@ -1,37 +1,102 @@
...
+
+HEADERS = [
+    fields.FIELD_UUID,
+    fields.FIELD_AIP_NAME,
+    fields.FIELD_CREATED_DATE,
+    fields.FIELD_AIP_SIZE,
+]
+
+
+def _prepare_csv_data(storage_service_id, rows, puids):
```
ablwr, line 20: Codecov is rightly interested in this. I guess if it doesn't work during the user testing phase, it'll be pretty obvious when it fails. But is there any kind of test coverage we can implement here to do checking around this function?

AIPscan/Reporter/report_aip_contents.py
```diff
@@ -1,37 +1,102 @@
...
+    query_results = (
+        db.session.query(db.func.count(File.file_format))
+        .join(AIP)
+        .join(StorageService)
+        .filter(StorageService.id == storage_service_id)
+        .filter(File.file_type == FileType.original.value)
+        .filter(AIP.uuid == aip_uuid)
+        .filter(File.puid == puid)
+    )
+    return str(query_results[0][0])
```
ablwr, line 74: If query_results[0] is null, how would this fail? Is that a risk?

AIPscan/Reporter/report_format_versions_count.py
```diff
@@ -1,39 +1,48 @@
...
     versions = version_data.get(fields.FIELD_FORMAT_VERSIONS)
 
-    headers = [
-        fields.FIELD_PUID,
-        fields.FIELD_FORMAT,
-        fields.FIELD_VERSION,
-        fields.FIELD_COUNT,
-        fields.FIELD_SIZE,
-    ]
+    headers = translate_headers(HEADERS)
```
ablwr, line 40: This is nice

AIPscan/Reporter/report_preservation_derivatives.py
```diff
@@ -0,0 +1,107 @@
...
+        {
+            fields.FIELD_AIP_UUID: file_[fields.FIELD_AIP_UUID],
+            fields.FIELD_AIP_NAME: file_[fields.FIELD_AIP_NAME],
+        }
+        for file_ in filtered_files
+    ]
+
+
+@reporter.route("preservation_derivatives/", methods=["GET"])
+def preservation_derivatives():
```
ablwr, line 74: This also looks like it could benefit from some extra code coverage.

ablwr (approved):
I have some small questions/comments,  mostly out of my own curiosity and being new to the codebase. Feel free to ignore. I don't see anything that concerns me. This is a great improvement to the overall healthiness of the code, with lots of tidy improvements and structural optimizations (in addition to the useful feature) to help us keep a consistent and clean home as this project grows.

AIPscan/Reporter/helpers.py
tw4l: I was thinking that we'd omit them because the primary keys might be noise/confusing to users - they're only internal to AIPscan, not visible in the UI, and don't reflect the data in the stored AIPs.

Re-import is an interesting question but I think for now we'll likely want to stick with fetching information from the Storage Service to ensure that stored AIPs and information in AIPscan don't diverge.

ablwr: OK, sounds good!

tw4l:
Thanks for the review and the encouragement to increase the code coverage @ablwr! We had very little coverage for the `Reporter` component to this point so I've added tests for all the new CSV views and some basic tests for the HTML tabular reports as well, which brought coverage for this PR from ~50% to ~92.5% and increased overall project coverage up 12.5% :smile:

ablwr:
Thanks for this great work!!

Issue #111, opened 2021-02-20
Problem: Fetch job hangs indefinitely if wrong SS URL or API key is used


peterVG:
Expected behaviour: AIPscan returns an error message that SS is not found or that SS authentication has failed and abandons the Fetch Job. No Fetch Job record is created in the DB.

Current behaviour: The AIPscan Fetch Job monitor window reports that downloading package lists returned a "FAILURE" but then proceeds to "Downloading AIP METS files" regardless. There are no further updates on the GUI. On the server side, AIPscan sends get requests to the SS Server (with missing parameter information that should have come from a successful package info response). The server runs these unresolved requests indefinitely. The only way to resolve is an AIPscan server reboot.

peterVG:
Error handling scaffolding for this has been provided by Ross here: https://github.com/artefactual-labs/AIPscan/blob/main/AIPscan/Aggregator/tasks.py#L182-L192

Need to determine:
1) how/where to log the error (i.e. in a log file, AIPscan db, or Celery db). Right now it's defaulting (successfully) to stderror.
2) whether to add error handling code (i.e. output error to user, stop SS request task, delete Fetch Job row from db) inside TaskError class (line 28) or in the error catching If statement (lines 187-192)

PR #128, opened 2021-04-27
Gracefully handle bad Storage Service connections

tw4l:
Connected to #111 

This PR checks the Storage Service credentials and ensures there is a good connection before kicking off a Fetch Job's Celery tasks. This prevents the most common reason for AIPscan's Celery tasks to hang and provides quick feedback in the UI if the Storage Service URL, username, or API key are incorrect, or if the Storage Service can't be reached:

image


ablwr (approved):
These are some great tests too!

PR #131, opened 2021-05-06
Fix reports table alignment

tw4l:
A very small CSS tweak to align the report selection table properly within the "Reports" page and stop Tessa from wincing every time she loads the page ;)

Before:

image

After:

image


ablwr:
YES absolutely

ablwr (approved):

Issue #87, opened 2020-10-28
Filter reports by storage location


tw4l:
At the October 22 sprint meeting, it was requested that reports should be filterable by storage location within a Storage Service. This might require extending the `StorageService` or `AIP` model to add the concept of storage location.

peterVG:
This new feature is very comprehensive and looking really good @tw4l nice work!

After a successful Fetch Job run, my AIPs and files can be filtered according to Storage Location from the Report dashboards page As well, given my 3 test Storage Locations, the "All locations" option reports the correct aggregate number of AIPs.

I like how if a Storage Location filter is selected, then that shows as part of the report header metadata:

![image](https://user-images.githubusercontent.com/672121/120859530-32bd3300-c539-11eb-9417-00dea9136c5e.png)

But if “All locations” is selected then the “Location” element is simply left out:

![image](https://user-images.githubusercontent.com/672121/120859631-54b6b580-c539-11eb-9f08-f4f414335548.png)

When we create proper user docs we should make note that that to maximize this feature's effectiveness, the Archivematica Admin should override the confusing action tense of the current default AM SS Location description and use something more descriptive.

For example, from this:
![image](https://user-images.githubusercontent.com/672121/120850642-8ffeb780-c52c-11eb-86b5-b0e1c02bcd8f.png)

To this:
<img width="845" alt="Screen Shot 2021-06-04 at 2 23 58 PM" src="https://user-images.githubusercontent.com/672121/120864111-8c752b80-c540-11eb-9d7e-15deae9e1af7.png">

Note that I had to completely destroy my AIPscan database to generate this ^ screencap.

I first renamed the awkwardly titled "Store AIP in standard Archivematica Directory" to "Default AIP Store" in the Storage Service:

<img width="1349" alt="Screen Shot 2021-06-04 at 1 06 36 PM" src="https://user-images.githubusercontent.com/672121/120858237-5f704b00-c537-11eb-8e74-65bdfdd94e30.png">

Then I deleted the 2 Fetch Jobs in my AIPscan database. I had zero Fetch Jobs. Then I started a new Fetch Job, expecting to pull in the 2 new Storage Location descriptions and the updated default storage location description. But this is what I saw instead.
<img width="1154" alt="Screen Shot 2021-06-04 at 1 06 15 PM" src="https://user-images.githubusercontent.com/672121/120858514-bd049780-c537-11eb-8527-ed20ccb86bb9.png">

[...]

PR #134, opened 2021-05-25
Add Storage Locations to AIPscan

tw4l:
Connected to #87 

This PR introduces a new `StorageLocation` model to AIPscan, which models an AIP storage location. Aggregator tasks are modified to write Storage Locations to the database and associate them with each AIP. All existing `Data` endpoints and reports are updated to include an option to filter by storage location, and the reports UI and individual reports are updated accordingly. Finally, a new "AIP storage locations" report is introduced which takes a Storage Service as input and lists details for each AIP store location, sorted descending by AIP count:

image

I've also taken the opportunity to do a bit of additional code cleanup as I went, fixing minor issues as I saw them.

Screenshot of updated reports UI (the Storage Service and Location dropdown filters are identical on the AIPs page):

image

New database ERD as of this work (I will update the associated issue after this is merged):

image


tw4l:
Hi @peterVG - this is now ready for review.

tw4l:
I've made an addition to add support for date range filtering on the Storage Service location report, using the AIP creation dates. This isn't a perfect metric but is consistent with how we're treating it in the rest of the AIPscan application so I think it makes sense given the client demand.

This is now ready for review @peterVG - thanks!

tw4l:
The latest commit better aligns the parameter selectors at the top of the Reports page, as per the now-closed https://github.com/artefactual-labs/AIPscan/issues/140: 

image

peterVG:
Date range filtering addition works well.

peterVG (approved):
Code is consistent and easy to read. Thanks for updating all the tests. They all pass in my env too.

PR #138, opened 2021-06-04
update tag line in header

peterVG:
The previous tagline used in both the GH repo and AIPscan app focussed on the "collect" phase but didn't mention the reporting phase of an AIPscan workflow. This is for historical reasons. Early designs were intended for AIPscan to serve purely as a data gathering and provisioning API. It added its own reporting modules in later rounds of development.

Lastly, the specific way that AIPscan "collects" its data could be called "crawling". While it leaves room for interpretation, this term gives the user a better mental image of what AIPscan actually does to "collect" its data.

### DIFF
--- a/AIPscan/templates/base.html
+++ b/AIPscan/templates/base.html
@@ -39,7 +39,7 @@
         </li>
       </ul>
     </div>
-  <span style="font-style: italic" class="navbar-text">Collect repository-wide info about Archivematica AIPs</span>
+  <span style="font-style: italic" class="navbar-text">Provide repository-wide reports by crawling Archivematica AIPs</span>
 </nav>
 
 <div class="container-fluid" style="margin: 20px 0;">


tw4l (commented):
Thanks Peter! This is a good change.

I'm wondering about the use of the word "crawl" though. To me that would imply that AIPscan is iterating over a bunch of AIPs (e.g. in a directory) to build its database, whereas really it's asking the Storage Service for METS files and then processing those.

How would you feel about something like "Repository-wide data and reporting on Archivematica AIPs" or even "Repository-wide Archivematica AIP reporting"?

peterVG:
I would like the tag line to communicate three things: (1) AIPscan "collects" AIP metadata. This is 50% of its purpose. (2) AIPscan "reports", that's the other half. (3) AIPscan is "repository-wide" for Archivematica AIPs.

Your revision suggestions go the opposite direction on (1), ie this functionality is not really indicated. 

I thought that "crawling" was descriptive and created a good mental model for non-technical users who don't care about the actual back-end implementation. What is "crawling" if not "requesting" a series of http-addressable resources and storing the response? 

Also, if the preference is to be more technically accurate than catchy, then using "AIP" is not really correct either. We are actually parsing METS files. 

I realize, however, that the common usage of the term is for public web page "crawling". As I mentioned in my earlier comment, it may not be 100% accurate but it's the closest I was able to get to a term that gives the right mental model. 

That is how I arrived at the term "scan" originally. 

This is my order of preference:

1) Crawl Archivematica AIPs to provide repository-wide reporting 

2) Repository-wide reporting from Archivematica AIPs

3) Provide repository-wide reports by scanning Archivematica AIPs

tw4l:
> I thought that "crawling" was descriptive and created a good mental model for non-technical users who don't care about the actual back-end implementation. What is "crawling" if not "requesting" a series of http-addressable resources and storing the response?

You know, that is a fair point! I think your first option "Crawl Archivematica AIPs to provide repository-wide reporting" is great.

tw4l (approved):

peterVG:
Damn. Sorry, I hit merge without realizing I had tweaked the tag line some more as a result of this convo but hadn't changed the commit...

PR #139, opened 2021-06-04
Return CSS class lost during CR

peterVG:
Accidentally removed a CSS class during CR cleanup. This prevents the Fetch stats from being hidden on initial page load. This commit returns the missing CSS class to fix that.


tw4l (changes requested):
Hey @peterVG! I think this branch might need to be rebased on main, or if it's simpler you could make a new branch off of current main and put the fix there. If merged as-is, we'll end up with some duplicate commits.

peterVG:
Thanks for catching that @tw4l It's only a tweak of a few characters so your suggestion makes most sense. Closing this PR.

Issue #146, opened 2021-06-07
Problem: detailed fetch stats section displays on Storage Service record page load


peterVG:
The stats should be collapsed and hidden. They appear when a user clicks on the "i" icon.

peterVG:
Fixed in #147

PR #147, opened 2021-06-08
Add CSS class to hide fetch job stats by default

peterVG:
fixes issue #146


tw4l (approved):
Looks good, thanks!

Only one minor thing - would you mind rewording the commit message to use sentence case? i.e. "Add" instead of "add". Here are the relevant guidelines for Archivematica, which we try to use consistently across projects: https://github.com/artefactual/archivematica/blob/stable/1.12.x/CONTRIBUTING.md#commit-messages. Thanks!

peterVG:
Absolutely, thanks for catching that and the link.

Issue #152, opened 2021-06-09
Problem: Date range filtering is not implemented for AIP contents report


peterVG:
Currently the report is generated for all AIPs in the database, regardless of any date range filters that might have been set by the user. EDIT: this looks to have been intentional but leaves the user wondering why date range works on similar reports but not this one. The AIP ingest date is the third column you see in the report. SInce its such a large report it would be useful to narrow down contents by these dates.

There is another usability issue related to this report as it has a very long horizontal scroll which is hard to read. 

This report also has an open print formatting bug: https://github.com/artefactual-labs/AIPscan/issues/150  

All things considered, I am actually not convinced that this report offers something better over the AIP contents browsing that is currently provided via the AIPscan GUI. The only thing it can't do right now is date range filtering (which is broken or not implemented for the report anyway).

Therefore, I would like to propose that we "hide" this report from the upcoming 0.5.0 release GUI rather than spend more time on it ATM.

tw4l:
The AIP Contents report was reworked in https://github.com/artefactual-labs/AIPscan/pull/154 and now has filtering by start and end date.

PR #154, opened 2021-06-09
Rework AIP Contents report

tw4l:
Connected to #152

Reworked print-friendly report (dropping the Datatable, so that all results are printable) in UI:

image

The CSV report uses a pipe (`|`) to separate the format strings, rather than a newline as in the web UI.

### DIFF
--- a/AIPscan/API/namespace_data.py
+++ b/AIPscan/API/namespace_data.py
@@ -12,7 +12,7 @@ from flask_restx import Namespace, Resource
 
 from AIPscan.API import fields
 from AIPscan.Data import data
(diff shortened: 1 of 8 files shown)


tw4l:
Hey @peterVG - here's a quick reworking of the AIP Contents report so that we don't have to drop it for 0.5.0. I think in its new form, the condensed form of this report has enough value to include it, and this PR also brings start and end date parameters to the AIP Contents report. There are still some reports that do not use those filters, but we're slowly working our way down the list.

peterVG (approved):
New date range test is passing. LGTM.

Issue #161, opened 2021-06-30
Problem: METS files are not deleted after they are parsed


tw4l:
This can cause disk full issues, as METS files accumulate on disk.

PR #162, opened 2021-06-30
Cleanup Fetch Job downloads directory

tw4l:
Connected to #161 

The `AIPscan/Aggregator/downloads/` directory was previously filling up during Fetch Jobs and causing disk space issues.

This PR modifies the `Aggregator` tasks to delete METS files and JSON package lists as soon as they are no longer needed.

Some empty directories are currently still left behind - e.g.:

```
AIPscan/Aggregator/downloads
└── 2021-06-30-12-14-35
    ├── mets
    │   ├── 1
    │   ├── 2
    │   ├── 3
    │   └── 4
    └── packages
```

Ideally we'd clean up these empty directories as well, but that will require larger changes to the asynchronous processing tasks. Until we can address that, it might be good to occasionally delete these empty directories through a cron job or similar.

### DIFF
--- a/AIPscan/Aggregator/database_helpers.py
+++ b/AIPscan/Aggregator/database_helpers.py
@@ -360,7 +360,7 @@ def collect_mets_agents(mets):
         if aip_file.use != ORIGINAL_OBJECT and aip_file.use != PRESERVATION_OBJECT:
             continue
         agents = agents + _get_unique_agents(aip_file.get_premis_agents(), agents)
-    logger.info("Total  agents: %d", len(agents))
+    logger.info("Total agents: %d", len(agents))
     return agents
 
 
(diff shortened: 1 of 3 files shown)


peterVG:
Screen Shot 2021-06-30 at 1 16 38 PM

LGTM. Watching METS files get "cleaned up" as the Fetch Job is happening ^

Confirmed that it leaves behind the empty Fetch Job directory and sub-directories. Deleting the Fetch Job from the GUI will delete the corresponding directory. But now that we have added the de-dupe feature it's actually preferred to keep your Fetch Job metadata in AIPscan so you have a log of Fetch updates.  Maybe could a note to docs somewhere about housekeeping the /downloads directory?

Thanks for tests. They passed for me.

peterVG (approved):

PR #163, opened 2021-07-05
Delete already-parsed METS files

tw4l:
Connected to #161 

This PR fixes an oversight whereby already-parsed METS files weren't being deleted after being downloaded.


peterVG (approved):
Works. Added two new AIPs to SS and re-ran Fetch Job. Watched /downloads folder as existing METS were skipped, when the two new METS files were downloaded, and then deleted.

Screen Shot 2021-07-05 at 6 25 01 PM

Updated test passes. LGTM!

PR #170, opened 2022-03-01
Dockerize AIPscan

sbreker:
Add the capbility to launch AIPscan using Docker. Three containers are
created: aipscan, rabbitmq, and celery-worker.

To use the default bridge network, AIPscan can be launched using the
default docker-compose.yml file:
`dc up -d`.

This will install AIPscan on a separate network from AM. In AIPscan the
storage service IP should be set to the host machine IP using the
forwarded port: e.g. 192.168.1.89:62081.

To attach to the default Docker AM network ('hack_default') use:
`dc -f docker-compose.yml -f docker-compose.am-network.yml up -d`

Using this will require that you configure AIPscan with the Storage
Service Docker container's assigned IP and port 8000. This IP can be
found using the command `docker network inspect hack_default`.


tw4l (approved):
Thanks @sbreker, this is a great and welcome improvement!

PR #171, opened 2022-04-12
Add start and end date parameters to user ingest log report and Data endpoint

tw4l:
Connected to #171

### DIFF
--- a/AIPscan/Data/report_data.py
+++ b/AIPscan/Data/report_data.py
@@ -430,10 +430,24 @@ def aips_by_puid(
     )
 
 
-def agents_transfers(storage_service_id, storage_location_id=None):
-    """Return information about agents involved in creating a transfer
-    and provide some simple statistics, e.g. ingest start time and
-    ingest finish time.
+def agents_transfers(
+    storage_service_id, start_date, end_date, storage_location_id=None
+):
+    """Return overview of transfers and their agents.
+
+    Includes simple statistics such as ingest start and finish times.
+
+    :param storage_service_id: Storage Service ID (int)
+    :param start_date: Inclusive AIP creation start date
+        (datetime.datetime object)
+    :param end_date: Inclusive AIP creation end date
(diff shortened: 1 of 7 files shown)


AIPscan/Data/report_data.py
```diff
@@ -430,10 +430,23 @@ def aips_by_puid(
...
+    :param storage_service_id: Storage Service ID (int)
+    :param start_date: Inclusive AIP creation start date
+        (datetime.datetime object)
+    :param end_date: Inclusive AIP creation end date
+        (datetime.datetime object)
+    :param storage_location_id: Storage Location ID (int)
+
+    :returns: "report" dict containing following fields:
+        report["StorageName"]: Name of Storage Service queried
+        report["FormatVersions"]: List of result files ordered desc by size
```
djjuhasz, line 449: I think this help text is incorrect, the returned fields are StorageName, StorageLocation, and a list of ingests (AIPUUID, User, IngestStartDate, IngestFinishDate).  I'm not sure if this is surfaced in the UI anywhere though, or just for developers.

AIPscan/Data/tests/test_agents_transfers.py
```diff
@@ -63,10 +90,18 @@ def test_agents_transfers(
```
djjuhasz, line 90: Looks like the "AIPName" field is returned from `report_data.agents_transfers()`, but missing from this test.

djjuhasz (commented):
Hi @tw4l, I added a couple of comments but neither are blockers.

AIPscan/Data/report_data.py
tw4l: you're totally right! lazy copy-paste error :woman_facepalming:

tw4l: it's also just for developers, fyi

djjuhasz (approved):
Looks good :)

PR #175, opened 2022-04-12
Add number of files to Storage locations report

tw4l:
Connected to #175

PR #177, opened 2022-04-13
Add storage location timeseries chart

tw4l:
Connected to #173

~~Rebase and merge only after https://github.com/artefactual-labs/AIPscan/pull/175, as this depends on that PR's commit, which is for now duplicated here.~~

This PR introduces a monthly timeseries chart version of the storage locations report. This chart shows monthly or cumulative totals for number of AIPs, total AIP size (in GB), and file counts, within the specified date ranges provided as parameters. By using helpers from pandas to turn the underlying dataframe into a CSV, we're also able to offer to a CSV download version of the data underpinning the chart through the "Download CSV" button in the report header.

image


AIPscan/Data/report_data.py
```diff
@@ -631,3 +634,110 @@ def storage_locations(storage_service_id, start_date, end_date):
...
+            # the same month.
+            local_start = datetime.strptime(month, "%Y-%m").replace(day=1)
+
+            if start_date.isoformat()[:7] == month and start_date > local_start:
+                local_start = start_date
+
+            if cumulative:
+                local_start = start_date
+
+            last_days_of_month = [31, 30, 29, 28]
```
djjuhasz, line 715: Opinion: I think this would be slightly more readable as a separate `get_last_day_of_month()` method.

AIPscan/Data/tests/test_storage_locations.py
```diff
@@ -179,3 +182,103 @@ def test_storage_locations_data(
...
+        start_date=parse_datetime_bound(DATE_BEFORE_AIP_1),
+        end_date=parse_datetime_bound(DATE_AFTER_AIP_3, upper=True),
+        cumulative=cumulative,
+    )
+
+    assert report[fields.FIELD_STORAGE_NAME] == "test storage service"
+
+    locations = report[fields.FIELD_LOCATIONS_BY_MONTH]
+
+    NUMBER_OF_MONTHS_IN_RANGE = 30
```
djjuhasz, line 239: I'm not sure why a variable assignment is necessary here?

AIPscan/Reporter/templates/report_storage_locations_timeseries.html
```diff
@@ -0,0 +1,60 @@
...
+  <span id="storageServiceID" style="display:none;">{{ storage_service_id }}</span>
+  <span id="startDate" style="display:none;">{{ start_date.strftime('%Y-%m-%d') }}</span>
+  <span id="endDate" style="display:none;">{{ end_date.strftime('%Y-%m-%d') }}</span>
+</div>
+
+<div class="noprint" style="margin-bottom: 10px;">
+  <strong>Metric:</strong>
+  <select id="metricSelector" class="form-control" style="width:auto;">
+  {% if metric == "size" %}
+    <option value="aips">AIPs</option>
```
djjuhasz, line 27: Opinion: I think "AIP count" is clearer.

AIPscan/Reporter/templates/report_storage_locations_timeseries.html
```diff
@@ -0,0 +1,60 @@
...
+    <option value="files">File count</option>
+  {% endif %}
+  </select>
+</div>
+
+<div class="noprint" style="margin-bottom: 10px;">
+  <strong>Show totals:</strong>
+  <select id="cumulativeSelector" class="form-control" style="width:auto;">
+  {% if cumulative %}
+    <option value="false">Monthly</option>
```
djjuhasz, line 46: Looking at the screenshot I assumed this select box changed the resolution of the graph (e.g. daily, monthly).  I think "Differential" or "Difference" would be clearer. Labelling the widget "Cumulative" with a checkbox for "Yes" could be another option.

djjuhasz (approved):
@tw4l I added a bunch of opinions, but nothing that's a blocker. :)

AIPscan/API/namespace_report_data.py
```diff
@@ -229,3 +229,42 @@ def get(self, storage_service_id):
...
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_END_DATE: {
+                "description": "AIP creation end date (inclusive, YYYY-MM-DD)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_CUMULATIVE: {
+                "description": "Return stats per month (False, default) or cumulative (True)",
```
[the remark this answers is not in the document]
tw4l: oh, nice! I agree :)

AIPscan/Data/report_data.py
```diff
@@ -631,3 +634,110 @@ def storage_locations(storage_service_id, start_date, end_date):
...
+
+            # Determine local start and end date values for the subquery.
+            # For start date, use first day of month or start_date, if later
+            # and in the same month. If cumulative, always use start date.
+            # For end date, sue last day of month or end_date, if prior and in
+            # the same month.
+            local_start = datetime.strptime(month, "%Y-%m").replace(day=1)
+
+            if start_date.isoformat()[:7] == month and start_date > local_start:
+                local_start = start_date
```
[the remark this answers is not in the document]
tw4l: Hm, I hear what you mean. Maybe I'll try setting the chart resolution per-day and see how manageable that is for large reporting periods (the client also expressed some interest being able to do this, though we both wondered if it would end up being too granular). Will try some things and then run it past you for a second opinion when I have something that seems like it's working well.

AIPscan/Data/tests/test_storage_locations.py
tw4l: hm, good point. maybe I was anticipating re-using this variable but never did. I'll get rid of it.

AIPscan/Data/report_data.py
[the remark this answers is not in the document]
tw4l: @djjuhasz commit https://github.com/artefactual-labs/AIPscan/pull/177/commits/2dbcdb7697f188b7324a8087b30999712cf6885c changes the reporting interval to daily - want to give it a spin and let me know what you think of the user experience? It does make the code a fair bit simpler in a good way, and I think will address your concern about user confusion.

I haven't updated the tests yet - figured I'd see what you think of the change first. Thanks!

djjuhasz:
@tw4l the daily time series report looks fine, but it does a little while to load (~3-4 seconds).  It doesn't seem to be limiting the  start and the end date of the report based on the first and last AIP dates either - but maybe that was a separate report? Also, when I tried to switch to the "cumulative" report I got an error.  See the attached screencast

https://user-images.githubusercontent.com/626135/167224859-49984a87-a564-44ca-87a4-a0bf11ee1267.mp4

.

tw4l:
@djjuhasz It should only limit the date range from the date the first AIP was stored to today if date parameters aren't passed in, or if invalid values are provided. You can test that by manually tweaking the start_date and end_date parameters in the URL :)

djjuhasz:
> @djjuhasz It should only limit the date range from the date the first AIP was stored to today if date parameters aren't passed in, or if invalid values are provided. You can test that by manually tweaking the start_date and end_date parameters in the URL :)

Ah, okay. 👍

Issue #174, opened 2022-04-12
Feature: Add new Largest AIPs report based on Largest files


tw4l:
~~Need to determine whether this would best be implemented as a new search parameter in the underlying data endpoint and Reports selection screen, or as a filter within the report.~~

Based on feedback from client, the feature request is to create a new Largest AIPs report similar to the existing Largest files report.

tw4l:
From client: "filter by AIP size, rather than file size."

PR #178, opened 2022-04-20
Add Largest AIPs report

tw4l:
Connected to #174

This PR adds a new Largest AIPs report to AIPscan, similar to the existing Largest files report.

image

Note: This PR includes a new `size` field on the `AIP` database model, which is a breaking change. Since we don't yet have database migrations in AIPscan, it is necessary to delete the existing `aipscan.db` SQLite database. When the application is restarted, a new database with the updated `AIP` model will be automatically created, and it will be necessary to run new Fetch Jobs to populate the database.

This PR also adds the date range parameters to the Largest files report, to make the user experience of these sibling reports consistent.

### DIFF
--- a/AIPscan/API/namespace_report_data.py
+++ b/AIPscan/API/namespace_report_data.py
@@ -108,6 +108,16 @@ class LargestFileList(Resource):
     @api.doc(
         "list_largest_files",
         params={
+            fields.FIELD_START_DATE: {
+                "description": "AIP creation start date (inclusive, YYYY-MM-DD)",
+                "in": "query",
+                "type": "str",
+            },
+            fields.FIELD_END_DATE: {
+                "description": "AIP creation end date (inclusive, YYYY-MM-DD)",
+                "in": "query",
+                "type": "str",
+            },
             fields.FIELD_FILE_TYPE: {
                 "description": "Optional file type filter (original or preservation)",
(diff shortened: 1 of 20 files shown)


AIPscan/API/namespace_report_data.py
```diff
@@ -56,11 +56,68 @@ def get(self, storage_service_id):
...
+        """List largest AIPs"""
+        start_date = parse_datetime_bound(request.args.get(fields.FIELD_START_DATE))
+        end_date = parse_datetime_bound(
+            request.args.get(fields.FIELD_END_DATE), upper=True
+        )
+        storage_location_id = request.args.get(fields.FIELD_STORAGE_LOCATION)
+        try:
+            limit = int(request.args.get(fields.FIELD_LIMIT, 20))
+        except ValueError:
+            pass
```
djjuhasz, line 96: Should limit be set to 20 if their is a ValueError?

djjuhasz (approved):
Looks good @tw4l.  One question, no blockers. :)

AIPscan/API/namespace_report_data.py
tw4l: yes, good point!

tw4l:
thanks @djjuhasz!

Issue #182, opened 2022-09-23
Bug: deleted AIPs statistic displays incorrect information


rayzilt:
"deleted AIPs" statistic under the information button is incorrect. The value is the same as AIP replicas

PR #183, opened 2022-09-23
Use correct variable at "deleted AIPs" statistic

rayzilt:
Variable for total_replicas is also used at deleted AIPs.

This pull request changes that variable to the correct one.

Issue: https://github.com/artefactual-labs/AIPscan/issues/182#issue-1383926163


mcantelon (approved):
Looks good to me! Sorry it took so long to review!

PR #185, opened 2023-01-13
Dev/add technical metadata v2

mcantelon:
Adds display of PREMIS object XML, if populated, when viewing files.

### DIFF
--- a/AIPscan/Aggregator/database_helpers.py
+++ b/AIPscan/Aggregator/database_helpers.py
@@ -5,6 +5,7 @@ database.
 """
 
 from celery.utils.log import get_task_logger
+from lxml import etree
 
 from AIPscan import db
 from AIPscan.Aggregator import tasks
@@ -57,8 +58,7 @@ def _extract_event_detail(premis_event, file_id):
 
 
 def _create_agent_type_id(identifier_type, identifier_value):
-    """Create a key-pair string for the linking_type_value in the db.
-    """
+    """Create a key-pair string for the linking_type_value in the db."""
     return "{}-{}".format(identifier_type, identifier_value)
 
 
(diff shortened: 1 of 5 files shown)


AIPscan/Aggregator/database_helpers.py
```diff
@@ -306,6 +307,18 @@ def _add_normalization_date(file_id):
...
+def _add_premis_object_xml(fs_entry, file_id):
+    """Add string representation of PREMIS Object to File object."""
+    file_ = File.query.get(file_id)
+    for ss in fs_entry.amdsecs[0].subsections:
+        if ss.contents.mdtype == fs_entry.PREMIS_OBJECT:
+            file_.premis_object = etree.tostring(
+                ss.contents.serialize(), pretty_print=True
+            )
+    db.session.commit()
+    return
```
Diogenesoftoronto, line 319: a return None is more explicit. Perhaps add return types and inputs in fields works fine either way though.

AIPscan/Aggregator/database_helpers.py
```diff
@@ -352,6 +365,8 @@ def create_file_object(file_type, fs_entry, aip_id):
     if file_type == FileType.preservation:
         _add_normalization_date(new_file.id)
 
+    _add_premis_object_xml(fs_entry, new_file.id)
+
```
Diogenesoftoronto, line 369: Maybe also add a return None here. I think you had the right idea in _add_premis_object_xml function to put a return statement. Even if it is implicit it is easier to read.

Diogenesoftoronto (commented):
I think what you have done is fine and in style with the rest of the repo. I don't think the changes I mentioned are that important they are just stylistic differences that might make more readable code. I think this is good to merge.

Diogenesoftoronto (approved):

PR #186, opened 2023-01-13
Change Flake repo in pre-commit.

mcantelon:
The previously used gitlab repo seems to be private now so switched to Flake's Github repo.


Diogenesoftoronto (approved):
all good pretty simple change

PR #187, opened 2023-01-17
Add Celery task for file cleanup after fetch job

mcantelon:
Adds a small Celery task to delete the directory in which files related to a fetch job are stored.


mcantelon:
There might be a better way to do this.

PR #190, opened 2023-01-20
Fix to PREMIS Object XML display.

mcantelon:
Changed storage of PREMIS object XML from binary to Unicode.

Added logic to make sure that there is PREMIS object XML before populating list of PREMIS object XML lines.


Diogenesoftoronto (approved):

Diogenesoftoronto (commented):
All good

PR #193, opened 2023-02-05
Add helper to update fetch job in DB.

mcantelon:
Added a helper to update a fetch job DB record with totals of different types of packages encountered during processing.


mcantelon:
I've added work from this PR to a different PR.

PR #194, opened 2023-02-06
Add helper to create fetch job in DB.

mcantelon:


mcantelon:
I've added work from this PR to a different PR.

PR #196, opened 2023-02-08
Add helpers to create/update fetch job in DB.

mcantelon:
Added fetch job-related helpers: one to create a new fetch job DB row and one to update an existing fetch job DB row with totals of different types pf packages encountered during processing.

### DIFF
--- a/AIPscan/Aggregator/database_helpers.py
+++ b/AIPscan/Aggregator/database_helpers.py
@@ -4,6 +4,8 @@
 database.
 """
 
+from datetime import datetime
+
 from celery.utils.log import get_task_logger
 from lxml import etree
 
@@ -18,6 +20,7 @@ from AIPscan.models import (
     Agent,
     Event,
     EventAgent,
+    FetchJob,
     File,
     FileType,
     Pipeline,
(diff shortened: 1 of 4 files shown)


AIPscan/Aggregator/database_helpers.py
```diff
@@ -402,3 +405,45 @@ def process_aip_data(aip, mets):
...
+    )
+    db.session.add(fetch_job)
+    db.session.commit()
+
+    return fetch_job
+
+
+def update_fetch_job(fetch_job_id, processed_packages, total_packages_count):
+    # Count different types of packages
+    total_aips = len([package for package in processed_packages if package.is_aip()])
```
Diogenesoftoronto, line 428: Doing this seems a little inefficient, you could instead have a single for loop it seems like and add to a count variable for each sip, dip and aip in processedpackage instead of making an array for each sip dip and aip and then checking the type.

total_sips
total_dips
total_aips
for package in processed_packages
    if package.is_aip() 
        total_aips++
    if package.is_dip()
        total_dips++
        ...

Diogenesoftoronto (changes requested):

Diogenesoftoronto (changes requested):
decent could just make it more efficient

AIPscan/Aggregator/database_helpers.py
mcantelon: True... and it's harder to read too.

mcantelon:
PR updated!

mcantelon:
Updated!

AIPscan/Aggregator/database_helpers.py
Diogenesoftoronto: I don't see the recommended change here. Should it not be implemented?

Diogenesoftoronto (changes requested):

mcantelon:
Updated it for real this time (I added the fix to the wrong branch)!

Diogenesoftoronto (approved):
awesome sauce

PR #197, opened 2023-02-14
Update Github workflow's actions to use Node16.

mcantelon:
See:

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


Diogenesoftoronto (approved):
Good to go

Issue #199, opened 2023-08-16
AIPScan database not updated if AIP deleted from storage service


aseles13:
**Current behaviour:** If AIPs are deleted from the storage service the AIPScan database will not change, meaning that deleted AIP is still appearing in search results. 
**Expected behaviour:** As AIPScan is a mechanism for searching and reporting against AIPs in the storage service if there changes that need to be made, they should be reflected in database results.
Related issues:
[Issue 91: Delete storage service doesn’t remove agents](https://github.com/artefactual-labs/AIPscan/issues/91)
[Issue 40: Delete fetch job times out](https://github.com/artefactual-labs/AIPscan/issues/40)

mcantelon:
Mel's fix for this has been merged!

PR #200, opened 2023-09-01
Update AIPScan database with deleted SS AIPs, #199

melaniekung:
[Related Issue: AIPScan database not updated if AIP deleted from storage service #199](https://github.com/artefactual-labs/AIPscan/issues/199)

When user runs "New Fetch Job", it will check each package to see if it's been deleted. If so, it will delete the aip from the aipscan.db.


mcantelon (approved):
Looks good (other than the nitpick)!

Issue #209, opened 2023-09-08
PROBLEM: AIPs page loads slowly


sarah-mason:
The AIPs view page can take several minutes to load in an instance, particularly those with a lot of AIPs in the database. No further searching or browsing of AIPs can take place at this time. If you click back to the AIPs page after viewing an AIP, it will take several minutes to load again.

A potential suggested fix was to add the search box to the main page where Fetch jobs are initiated to prevent the need to load the entire AIP view. There may also be other database optimisation options.

mcantelon:
Thanks Sarah... we'll take a look at potential solutions!

mcantelon:
I think I might know the underlying issue and am going to start on a fix.

mcantelon:
A pull request has been created to fix this: https://github.com/artefactual-labs/AIPscan/pull/215

I'll update this issue when it's reviewed and merged.

mcantelon:
I've merged a fix for this.

PR #215, opened 2023-09-28
Use server-side paging in aips page. (#209)

mcantelon:
To avoid slow page responses when there are a large amount of AIPs use server-side paging to limit the number of AIP HTML sent to the browser.

### DIFF
--- a/AIPscan/Reporter/helpers.py
+++ b/AIPscan/Reporter/helpers.py
@@ -145,3 +145,14 @@ def get_premis_xml_lines(file_object):
         premis_xml_lines = file_object.premis_object.split("\n")
 
     return premis_xml_lines
+
+
+def calculate_paging_window(pagination):
+    first_item = ((pagination.page - 1) * pagination.per_page) + 1
+    last_item = pagination.page * pagination.per_page
+
+    return first_item, min(last_item, pagination.total)
+
+
+def remove_dict_none_values(values):
+    return {index: "" if value is None else value for (index, value) in values.items()}
--- a/AIPscan/Reporter/views.py
+++ b/AIPscan/Reporter/views.py
@@ -39,7 +39,11 @@ from AIPscan.Reporter import (  # noqa: F401
     request_params,
     sort_puids,
 )
-from AIPscan.Reporter.helpers import get_premis_xml_lines
+from AIPscan.Reporter.helpers import (
+    calculate_paging_window,
+    get_premis_xml_lines,
+    remove_dict_none_values,
+)
 
 
 def _get_storage_service(storage_service_id):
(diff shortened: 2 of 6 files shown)


AIPscan/Reporter/helpers.py
```diff
@@ -145,3 +146,21 @@ def get_premis_xml_lines(file_object):
...
+
+    return first_item, last_item
+
+
+def urlencode_without_none_values(values):
+    for key in values:
+        if values[key] is None:
+            values[key] = ""
+
+    return urlencode(values)
```
sevein, line 166: Maybe a good use case for `values.items()`, e.g.:

```python
def urlencode_without_none_values(values):
    """URL encode the values dict converting any None values to empty strings."""
    for key, value in values.items():
        if value is None:
            values[key] = ""
    return urlencode(values)
```

Here's an alternative using a dictionary comprenhension, I think this can be useful if you want to avoid mutating the input dictionary or doing it during iteration.

```python
def urlencode_without_none_values(values):
    """URL encode the values dict converting any None values to empty strings."""
    return urlencode(
        {
            key: (value if value is not None else '')
            for key, value in values.items()
        }
    )
```

AIPscan/Reporter/helpers.py
```diff
@@ -145,3 +146,21 @@ def get_premis_xml_lines(file_object):
...
 
     return premis_xml_lines
+
+
+def calculate_paging_window(page, pagination):
+    first_item = ((page - 1) * pagination.per_page) + 1
+    last_item = page * pagination.per_page
+
+    if last_item > pagination.total:
+        last_item = pagination.total
```
sevein, line 156: An alternative to calculate last_item that is more concise by using `min()`:

```suggestion
    last_item = min(page * pagination.per_page, pagination.total)
```

AIPscan/Reporter/views.py
```diff
@@ -85,10 +89,24 @@ def view_aips():
     except Exception as e:
         print(e)
 
-    aips = AIP.query.filter_by(storage_service_id=storage_service.id)
+    page = int(request.args.get(request_params.PAGE, default="1"))
```
sevein, line 92: Should we handle `ValueError` when int doesn't receive a valid base 10 literal?

sevein (changes requested):
Hi @mcantelon! This looks good to me. I  haven't tested but I can see that what you're doing is going to bring a big improvement. I've made some small suggestions, but nothing important functionally speaking. Thank you!

mcantelon:
Thanks @sevein !

AIPscan/Reporter/helpers.py
mcantelon: Very handy!

mcantelon:
I've updated the PR @sevein .... thanks for the tips!

sevein (approved):
LGTM

Issue #216, opened 2023-09-30
AIP page doesn't use server-side paging


mcantelon:
The page that lists the comments of an AIP (`reporter/aip/<aip ID>`) uses client-side paging rather than server-side paging. This means that HTML representing every file in the AIP ends up getting sent to the end user's browser, which potential leads to very slow page loads. Paging should be changed so it's server-side rather than client-side.

mcantelon:
This will reuse helpers in the PR for adding server-side paging to the AIPS page.

It'll require:
* Changing AIPscan/Reporter/templates/aip.html to use server-side paging
* Changing view_aip in AIPscan/Reporter/views.py to use server-side paging

mcantelon:
A pull request has been created to address this issue: https://github.com/artefactual-labs/AIPscan/pull/221

mcantelon:
This work's now been merged.

PR #221, opened 2023-10-11
Use server-side paging in aip files page. (#216)

mcantelon:
To avoid slow page responses, when there are a large amount of files in an AIP, use server-side paging to limit the amount of AIP file HTML sent to the browser.


sevein (approved):
Looks good to me!

mcantelon:
Thanks @sevein !

PR #233, opened 2023-10-21
Fix errors when database is unpopulated (#234)

mcantelon:
Fixed errors that would result when the database was unpopulated. Previously the "AIPs" and "Reports" pages would result in an "Internal Server Error".


mcantelon:
I'm going to try a different approach.

mcantelon:
Re-opening: the other approach seems complicated.

Issue #102, opened 2020-12-14
It is difficult to isolate Artefactual code from vendor libraries


ross-spencer:
07d46ded introduced a vendor library for Font Awesome. This helpfully allows us to tease apart our code files vs. those of fontawesome. We still have other vendor libraries and so if we want to do something specific with our own, e.g. linting, then the exceptions needed in automation  scripts get more complicated. 

Consider moving everything to their own vendor library now as well as introduce vendor folders for new code while we have a good opportunity in the early phases of this project.

ross-spencer:
Possibly related? https://github.com/artefactual-labs/AIPscan/issues/82

mcantelon:
Merged PR to bundle assets using Vite build.

PR #236, opened 2023-10-29
Use Webpack to bundle vendor assets (#102)

mcantelon:


mcantelon:
I'm going to use vite for bundling.

Issue #230, opened 2023-10-20
Add tests to API blueprint


mcantelon:
The API blueprint currently has no tests. It would be good to add them to avoid regressions and be able to reference to get a quick sense of API functionality.

Tests can apparently be created the same way other tests of Flask endpoints are (see: https://github.com/python-restx/flask-restx/issues/189).

mcantelon:
Tests for these are actually in the Data blueprint.

PR #237, opened 2023-11-01
Add CLI tool to fetch AIPs (#230)

mcantelon:
Added a tool to fetch AIPs, either all at once or in "pages".

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,49 @@ adminsitrators.
 The test data generator, `tools/generate-test-data`, tool populates
 AIPscan's databse with randomly generated example data.
 
+#### Fetch script
+
+The AIP fetch tool, `tools/fetch_aips`, allows all, or a subset, of a storage
+service's packages to be fetched by AIPscan. Any AIPs not yet fetched by
+AIPscan will be added but no duplicates will be added if an AIP has already
+been fetched. Any AIPs that have been newly marked as deleted will be removed
+from AIPscan.
+
+When using the script the storage service's list of packages can optionally be
+grouped into "pages" with each "page" containing a number of packages
+(specified by a command-line argument). So, for example, packages on a storage
+service with 150 packages on it could be fetched by fetching three pages of 50
+packages. Likewise if the storage service has anything from 101 to 149 packages
+on it it could also be fetched by fetching three pages of 50 packages.
+
+If using `cron`, or some other scheduler, to automatically fetch AIPs using
+this tool consider using the `flock` command to prevent overlapping executions
+of the tool.
+
+##### Cached package list
+
+A storage service's list of packages is downloaded by the script and is cached
+so paging, if used, will remain consistent between script runs. The cache of a
+particular cached list of packages is identified by a "session descriptor". A
+session descriptor is specified by whoever runs the script and can be any
+alphanumeric identifier without spaces or special characters. It's used to name
+the directory in which fetch-related files are created.
+
+Below is what the directory structure would end up looking like if the session
+identifier "somedescriptor" was used, showing where the `packages.json` file,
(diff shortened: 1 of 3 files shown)


tools/fetch_aips
```diff
@@ -0,0 +1,140 @@
...
+        format="%(asctime)s:%(levelname)s:%(name)s:%(message)s",
+        filename=logfile,
+    )
+
+    logger_name = os.path.basename(sys.argv[0])
+    logger = logging.getLogger(logger_name)
+    logger.setLevel(logging.INFO)
+
+    root = logging.getLogger()
+    root.setLevel(logging.INFO)
```
sevein, line 45: At some point we should have a `get_logger` or similar for consistent logging across multiple tools.

tools/fetch_aips
```diff
@@ -0,0 +1,140 @@
...
+            end_item = total_packages
+
+        # Make sure paging window is valid and delete fetch job if not
+        if start_item > total_packages:
+            logger.error(
+                f"Fetch job deleted. Page {page} would start at package {start_item} but there are only {total_packages} packages."
+            )
+            db.session.delete(fetch_job)
+            db.session.commit()
+            sys.exit(1)
```
sevein, line 99: I don't know how it's supposed to be done in click but looking at the docs (https://click.palletsprojects.com/en/8.1.x/exceptions/), this seems like a better option?

    err = click.ClickException("error message")
    err.exit_code = 1
    raise err

tools/fetch_aips
```diff
@@ -0,0 +1,140 @@
...
+        )
+        logger.info("%s", summary)
+        logger.info(
+            f"Updated fetch job record {fetch_job_id} with package type counts."
+        )
+        logger.info("Processing complete.")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
sevein, line 140: Looks like main is doing some extra work because it's a click.command, seehttps://click.palletsprojects.com/en/8.1.x/exceptions/. It looks from some examples I've seen that `sys.exit` might not be necessary in this context.

sevein (approved):
Nicely done. Made some minor suggestions. Also I'd like to share a few thoughts for future reference. These points aren't urgent:

1. I think that we could try to integrate crucial business logic like fetching AIPS more closely with the core application modules (AIPScan package), closer to the models, service, etc... The CLI can then act as a simple interface to these functionalities.
2. Considering the significance of the feature it would be useful at some point ot develop an integration test for the CLI command, but I realize this could extend the dev time significantly.
3. For file system interactions, I'd recomment to adopt `pathlib` going forward since it's becoming the standard in modern Python.

README.md
```diff
@@ -165,6 +165,49 @@ adminsitrators.
...
+When using the script the storage service's list of packages can optionally be
+grouped into "pages" with each "page" containing a number of packages
+(specified by a command-line argument). So, for example, packages on a storage
+service with 150 packages on it could be fetched by fetching three pages of 50
+packages. Likewise if the storage service has anything from 101 to 149 packages
+on it it could also be fetched by fetching three pages of 50 packages.
+
+If using `cron`, or some other scheduler, to automatically fetch AIPs using
+this tool consider using the `flock` command to prevent overlapping executions
+of the tool.
```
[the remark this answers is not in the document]
mcantelon: Good thinking! I've added an issue for this.

tools/fetch_aips
mcantelon: Sounds good! I've added an issue to add this.

mcantelon:
Thanks for the review @sevein ! I've addressed the small issues and added Github issues for the various recommendations as they all make sense.

PR #238, opened 2023-11-07
Use Vite to bundle vendor assets (#102)

mcantelon:
* Added Vite build bundling of JavaScript
* Removed unused DataTable-related JavaScript
* Updated JavaScript libraries


Makefile
```diff
@@ -17,3 +17,18 @@ schema-docs:
...
 		-debug
+
+build:
+	# Install Node modules if not present
+	if [ ! -d "node_modules" ]; then \
+		npm install; \
+	fi
+
+	# Bundle vendor assets
+	vite build -c build_config/base.config.js
```
sevein, line 28: We should probably use `npx vite` instead to ensure that it's using the vite installed in this environment.

Makefile
```diff
@@ -17,3 +17,18 @@ schema-docs:
 		-cat aipscan \
 		-s aipscan \
 		-debug
+
+build:
+	# Install Node modules if not present
+	if [ ! -d "node_modules" ]; then \
+		npm install; \
+	fi
```
sevein, line 25: Wouldn't you want to always run "npm install" or "npm ci"?

sevein (approved):
It builds really fast. I've made a couple of minor suggestions.

The size of the PR makes my browser very unhappy, hard to browse the files. There must be something going on with GitHub's UI.

Makefile
mcantelon: Good call!

Makefile
mcantelon: Good point... npm doesn't re-install what's already installed.

Issue #248, opened 2023-11-14
Add a script for the updating of ERD 


aseles13:
This is related to closed issue #42

mcantelon:
PR merged!

PR #249, opened 2023-11-15
Add script to generate DB documentation (#248)

mcantelon:
Added a Bash script to generate database documentation using SchemaSpy. The script requires that Graphviz and a Java Runtime Environment be intstalled.


sevein (approved):
Looks great, didn't know the tool.
If you're not purposely avoiding Docker, maybe their public image would be more convenient: https://hub.docker.com/r/schemaspy/schemaspy/.

mcantelon:
Thanks @sevein ! Oh yeah, forget there might be a docker image for that. That could be simpler.

mcantelon:
Ended up implementing this a different way.

Issue #23, opened 2020-07-03
window.location.origin paths will not work on sites where AIPscan is not hosted at root url


peterVG:
window.location.origin values are currently used on `reporter/reports/` and `reporter/view_aips`. The urls they generate will fail if AIPscan is being hosted at sub-directories on a root url, e.g. `example.com/dir/AIPscan/` instead of `example.com/AIPscan`.

mcantelon:
Use of `window.location.origin` occurs in:
* Reporter/templates/aips.html (1 occurance)
* Reporter/templates/reports.html (17 occurances)
* static/js/report.js (4 occurrances), used by templates/report_base.html

The Flask `request` object specifies the app root as `url_root` so we can make this accessible to JavaScript via a data attribute in the `base.html` and `report_base.html` templates like:

    <body data-url-root="{{ request.url_root }}">

We can then retrieve it from JavaScript using `$('body').data('url-root')`.

mcantelon:
Fix merged.

PR #252, opened 2023-11-16
Fix JavaScript base URLs (#23)

mcantelon:
Removed JavaScript use of window.origin and window.location.origin to get application base URL. Changed to allow Flask to provde the base URL. Also did some minor code cleanup.


sevein (approved):
LGTM

mcantelon:
Thanks @sevein!

Issue #28, opened 2020-07-08
Pylint: Replace built-in names with different choices for variable name


ross-spencer:
I'll post a couple of high priority pylint fixes to address. The first are the use of reserved names for other variables. Unfortunately nice-looking names, but there are other patterns we can adopt here that will help make the code more readable:

From `pylint`:
```
Pylint priorities - really subtle undefined behavior for these:

[W0622 redefined-builtin] Redefining built-in 'list' File: Aggregator/tasks.py, line 87, in workflow_coordinator
[W0622 redefined-builtin] Redefining built-in 'next' File: Aggregator/tasks.py, line 198, in package_lists_request
[W0622 redefined-builtin] Redefining built-in 'file' File: Aggregator/tasks.py, line 268, in get_mets
[W0622 redefined-builtin] Redefining built-in 'type' File: Aggregator/tasks.py, line 348, in get_mets
[W0622 redefined-builtin] Redefining built-in 'format' File: Aggregator/tasks.py, line 317, in get_mets
[W0622 redefined-builtin] Redefining built-in 'id' File: Aggregator/views.py, line 38, in storage_service
[W0622 redefined-builtin] Redefining built-in 'id' File: Aggregator/views.py, line 55, in edit_storage_service
[W0622 redefined-builtin] Redefining built-in 'id' File: Aggregator/views.py, line 109, in delete_storage_service
[W0622 redefined-builtin] Redefining built-in 'id' File: Aggregator/views.py, line 123, in new_fetch_job
[W0622 redefined-builtin] Redefining built-in 'id' File: Aggregator/views.py, line 197, in delete_fetch_job
[W0622 redefined-builtin] Redefining built-in 'format' File: models.py, line 131, in originals.__init__
[W0622 redefined-builtin] Redefining built-in 'format' File: models.py, line 170, in copies.__init__
[W0622 redefined-builtin] Redefining built-in 'type' File: models.py, line 212, in events.__init__
[W0622 redefined-builtin] Redefining built-in 'type' File: models.py, line 231, in agents.__init__
[W0622 redefined-builtin] Redefining built-in 'id' File: Reporter/views.py, line 22, in view_aips
[W0622 redefined-builtin] Redefining built-in 'id' File: Reporter/views.py, line 56, in view_aip
[W0622 redefined-builtin] Redefining built-in 'id' File: Reporter/views.py, line 86, in view_original
[...]

PR #256, opened 2023-11-19
Add check for Python built-ins, remove uses (#28)

mcantelon:
Removed uses of Python built-ins and added a CI check for their
presence.


mcantelon:
The flake-builtins plugin seems to alert to some false positions.

https://github.com/semgrep/bento/issues/197

Issue #149, opened 2021-06-08
Problem: CSV exports are not sortable by size column


tw4l:
See related Archivematica issue: https://github.com/archivematica/Issues/issues/1450

This was fixed for the CSV download on Archivematica's Archival Storage tab by adding an additional sortable "Bytes" column to the CSV.

mcantelon:
The size data gets formatted by the `format_size_for_csv` function in `AIPscan/Reporter/helpers.py`. We could possible add logic to this to auto-insert the "Bytes" column.

mcantelon:
I've created a PR to fix this issue.

https://github.com/artefactual-labs/AIPscan/pull/261

mcantelon:
Merged fix.

PR #260, opened 2023-12-01
Add sortable size CSV field (#149)

mcantelon:


mcantelon:
This isn't a good way to do this, so closing.

PR #261, opened 2023-12-02
Add sortable size CSV field (#149)

mcantelon:
Package and file size values in CSV exports were made to be human readable which was good for legibility, but impeded the ability to sort rows by size. Added automatic population of, whenever size data is included in an export, an additional column containing the "raw" byte value of a package or file size.


sevein (approved):
LGTM

Issue #243, opened 2023-11-07
Unify CLI fetch tool logic and main app logic


mcantelon:
The CLI fetch tool logic is replicates the main app's fetch logic to a large degree so we should unify the code with module(s) in the main app.

mcantelon:
Merged some work to improve this (but there's still some work to be done).

mcantelon:
A PR for the remaining work has been created: https://github.com/artefactual-labs/AIPscan/pull/299

mcantelon:
Merged PR. @sevein had a good idea, though. He could have the CLI fetch tool make a REST API request to trigger fetching rather than having independent logic.

PR #263, opened 2023-12-05
Code cleanup/refactoring (#243)

mcantelon:
* Replaced use of api_url dict with StorageService instance or ID
* Did some refactoring to unify the AIP fetching code between the main app logic and the CLI tool


sevein (approved):
I haven't tested but LGTM. I've noticed that the SS client logic coold be improved too, e.g. we're dealing with HTTP transport concerns in modules like database_helpers, etc. That could use some refactoring too in the future, e.g. a dedicated module with a Client or similar, something that could eventually be extracted into its own project if we wanted.

mcantelon:
Thanks @sevein ! Yeah, agreed that further refactoring would definitely help. This is kind of a first pass.

PR #277, opened 2024-01-12
Add mysqlclient python dependency

scollazo:
This allows for the use of mysql as backend.

SQLALCHEMY_DATABASE_URI environment variable needs to be defined with the mysql db, with format "mysql://user:pass@host/db"


scollazo:
Changed from mysqlclient (c bindings) , to pymsql (native bindings), to avoid dependency on gcc

mcantelon (approved):
This looks great @scollazo ! Thanks!

scollazo:
Merged as https://github.com/artefactual-labs/AIPscan/commit/ddaaf1c7c409903bdd227f9b4e29493524698602, but I did something wrong and the pr doesn't autoclose

PR #284, opened 2024-01-23
Add initial documentation page

sallain:
Add a docs directory and an initial page for usage. I've added a couple headings to the page that I will fill in later.


mcantelon (approved):
Seems good!

Issue #287, opened 2024-01-25
The AIP view page doesn't handle non-existent AIP IDs gracefully


mcantelon:
Visiting the AIP view page using a non-existent AIP ID - /reporter/aip/122222222222222222222222222222222222 for example - results in an internal server error rather than a normal 404 page.

mcantelon:
These views don't behave correctly either:

No error (should 404): http://127.0.0.1:5000/aggregator/storage_service/100000000000000000

Error (should 404): http://127.0.0.1:5000/reporter/file/100000000000000000000000000

mcantelon:
Fix PR was merged.

PR #288, opened 2024-01-25
Return HTTP 404 if view resource not found (#287)

mcantelon:
Added 404 responses to storage service view, storage service edit view, storage service delete view, fetch job delete view, AIP view, and file view.


sevein (approved):
:+1:

Issue #297, opened 2024-03-19
Use real format types in fake data generation


mcantelon:
Use real format PUIDs, format names, versions, etc. in the fake data generator.

mcantelon:
PR ready for review: https://github.com/artefactual-labs/AIPscan/pull/298

PR #298, opened 2024-03-19
Use real format types in data generator (#297)

mcantelon:
Use real format types, read from a CSV file, in the fake data generation CLI tool. Allow the number of possible format types, from the CSV file, to be specified using a command-line option.


mcantelon:
Thanks @sevein !

PR #299, opened 2024-03-19
Refactor fetch CLI tool logic (#243)

mcantelon:
Refactor fetch CLI tool, sharing more functionality with the fetch Celery task.


sevein (approved):
LGTM! Perhaps another approach that we could investigate in the future is to have the CLI perform as an API client.

mcantelon:
Thanks @sevein !

(And good idea having the CLI call an API. I'll make an issue to look into that!)

Issue #335, opened 2024-07-01
Add search to AIPs page


mcantelon:

mcantelon:
PR for this was merged: https://github.com/artefactual-labs/AIPscan/commit/c8592bd3b61c49121250675f2d87949bf97adf60

PR #336, opened 2024-07-03
Add AIPs search (#335)

mcantelon:
* Added text search to AIPs page
* Cleaned up AIP paging test

### DIFF
--- a/AIPscan/Reporter/views.py
+++ b/AIPscan/Reporter/views.py
@@ -86,24 +86,35 @@ def storage_locations_with_aips(storage_locations):
     return [loc for loc in storage_locations if loc.aips]
 
 
-def get_aip_pager(page, per_page, storage_service, storage_location):
-    try:
-        page = int(page)
-    except ValueError:
-        page = 1
-
+def get_aip_pager(page, per_page, storage_service, **kwargs):
     storage_service_id = None
     if storage_service is not None:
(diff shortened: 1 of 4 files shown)


AIPscan/Reporter/views.py
```diff
@@ -86,25 +86,36 @@ def storage_locations_with_aips(storage_locations):
...
+    if storage_location_id:
+        aips = aips.filter_by(storage_location_id=storage_location_id)
+
+    # Optionally filter by text search
+    if query is not None:
+        aips = aips.filter(
+            AIP.uuid.like(f"%{query}%")
+            | AIP.transfer_name.like(f"%{query}%")
+            | AIP.create_date.like(f"%{query}%")
         )
```
sevein, line 115: This guarantees a full table scan, it won't be very efficient. I guess that's ok for now?

sevein (approved):

AIPscan/Reporter/views.py
mcantelon: Yeah, this is just to get something in place to replace functionality that seemingly once existed (a client had made their own documentation of it with screenshots), yet weirdly doesn't seem to exist in the code (I scoured the commit history and could find any trade). Very weird!

If it's slow for them I might re-implement it with Typesense.

mcantelon:
Thanks @sevein !

Issue #342, opened 2024-09-19
Typesense version of format count report generates 500 error when there are no results


mcantelon:
The Typesense version of format count report  (`/reporter/report_formats_count`) generates a 500 error for queries with no results.

mcantelon:
PR for CR: https://github.com/artefactual-labs/AIPscan/pull/343

mcantelon:
Fix merged.

PR #343, opened 2024-09-19
Fix format-related Typesense report errors (#342)

mcantelon:
Don't bother doing a counts of formats, and format versions, of files matching report criteria if no formats, or format versions, were found (given that otherwise a multi-search Typesense query will be sent with no searches and will result in an HTTP 500 error).


djjuhasz (approved):
LGTM :+1:

PR #378, opened 2025-05-07
Replace Python 3.8 support with 3.11/3.12 support

mcantelon:
Python 3.8's EOL was 2024-10-07. Removed and added Python 3.11 and 3.12 support making Python support the same as Archivematica's.

https://devguide.python.org/versions/

Also removed use of distutils module, depricated in Python 3.10.


sevein (approved):
:+1:

mcantelon:
Thanks @sevein !

Issue #379, opened 2025-05-07
When a storage service is deleted, it still shows as existing on the response page


mcantelon:

mcantelon:
PR for CR: https://github.com/artefactual-labs/AIPscan/pull/380

PR #380, opened 2025-05-07
Lessen display of deleted storage services (#379)

mcantelon:
When a storage service was deleted by a user the response page would, because the deletion takes place asynchronously, show the storage service in the list of storage services. Changed the behavior to not redirect immediately back to the list of storage services.


djjuhasz (approved):
LGTM @mcantelon. :+1:

mcantelon:
Thanks @djjuhasz !

PR #382, opened 2025-05-08
Fix web UI issue with empty fetches (#381)

mcantelon:
This fixes an issue where an attempted fetch, from a storage service containing no non-deleted AIPs, would never show as completed in the web UI.


djjuhasz (approved):
LGTM :+1:

mcantelon:
Thanks @djjuhasz !

PR #386, opened 2025-09-10
Improve MySQL compatibility

sevein:
This pull request enhances MySQL compatibility in AIPscan by introducing database schema adjustments, application logic refinements, and updates to the development environment to make use of it as the default database engine.

It fixes (82f19ca24d03d9cfd503973700beeb9d4421297c) an issue where fetch jobs could become stuck in an infinite loop because the expected task record was never visible to the session under MySQL’s REPEATABLE READ isolation level. The update adjusts how task visibility is handled so that dependent tasks can be detected reliably, preventing fetch jobs from hanging indefinitely.


docker-compose.yml
```diff
@@ -2,18 +2,56 @@ volumes:
...
+      - SQLALCHEMY_CELERY_BACKEND=mysql+pymysql://aipscan:demo@aipscan-mysql:3306/aipscan
+      - CELERY_RESULT_BACKEND=db+mysql+pymysql://aipscan:demo@aipscan-mysql:3306/celery
       - TYPESENSE_API_KEY=xyz
+      - CREATE_DB=true
     ports:
       - 5000:5000
+    command: gunicorn --preload -t 300 --workers 3 --bind 0.0.0.0:5000 "AIPscan:create_app()"
     volumes:
       - "./:/src:rw"
+        #- "/tmp/data:/src/AIPscan/Aggregator/downloads/"
```
replaceafill, line 42: Is this needed?

AIPscan/Aggregator/views.py
```diff
@@ -228,29 +229,12 @@ def new_fetch_job(fetch_job_id):
...
-    while True:
-        obj = package_tasks.query.filter_by(workflow_coordinator_id=task.id).first()
-        try:
-            task_id = obj.package_task_id
-            if task_id is not None:
-                break
-        except AttributeError:
-            continue
+    fetch_job_id = fetch_job.id
+    task_id = _wait_for_package_list_task_id(task.id)
```
replaceafill, line 233: Would it be a problem If `task_id` becomes `None` here?

AIPscan/models.py
```diff
@@ -398,7 +399,7 @@ class File(db.Model):
     format_version = db.Column(db.String(255))
     checksum_type = db.Column(db.String(255))
     checksum_value = db.Column(db.String(255), index=True)
-    premis_object = db.Column(db.Text())
+    premis_object = db.Column(LONGTEXT)
```
replaceafill, line 402: I think this is the reason [tests are currently failing](https://github.com/artefactual-labs/AIPscan/actions/runs/17618665997/job/50058456352) and this needs to support `sqlite` as well?

docker-compose.yml
```diff
@@ -2,18 +2,56 @@ volumes:
...
   aipscan:
     build: ./
     environment:
       - CELERY_BROKER_URL=amqp://guest@rabbitmq//
       - TYPESENSE_HOST=typesense
+      - SQLALCHEMY_DATABASE_URI=mysql+pymysql://aipscan:demo@aipscan-mysql:3306/aipscan
+      - SQLALCHEMY_CELERY_BACKEND=mysql+pymysql://aipscan:demo@aipscan-mysql:3306/aipscan
+      - CELERY_RESULT_BACKEND=db+mysql+pymysql://aipscan:demo@aipscan-mysql:3306/celery
       - TYPESENSE_API_KEY=xyz
+      - CREATE_DB=true
```
replaceafill, line 36: Is there any scenario where this needs to be `false`?

replaceafill (changes requested):
@sevein @scollazo This looks good overall. I appreciate the updates to the `docker-compose.yml` file :bow: I left a few minor questions, but I think the MySQL specific dialect used for one field breaks `sqlite`.

PR #387, opened 2025-09-13
Fix security vulnerabilities by updating dependencies

sevein:
This is not a general upgrade of all dependencies, only the bare minimum required to resolve known vulnerabilities.


replaceafill (approved):
:+1:

PR #391, opened 2025-09-18
Use Python 3.13

sevein:
This commit also updates pandas from 1.3.5 to 2.3.2 and statsmodels from 0.14.1 to 0.14.5 for better compatibility with recent versions of Python. This should speed up build times too, i.e. it can take near 10m now for some versions of Python.

image


replaceafill (approved):
Thanks @sevein! I really appreciate the SQLAlchemy migration :bow:

PR #397, opened 2025-09-29
Switch release Docker to standard Python and pip

sevein:
This pull request refactors the release container image to use the standard Python distribution instead of Astral's standalone Python builds, adopting a more conservative approach aligned with what we've been deploying in other environments. It also adds support for linux/arm64 builds.


Dockerfile
```diff
@@ -36,19 +40,14 @@ WORKDIR /app
...
-ENV UV_LINK_MODE=copy
-ENV UV_PYTHON_PREFERENCE=only-managed
-RUN --mount=type=bind,source=.python-version,target=/.python-version,ro \
-    uv python install
-RUN uv venv /venv
-ENV PATH="/venv/bin:$PATH"
-RUN --mount=type=cache,target=/root/.cache/uv \
-    --mount=type=bind,source=dist,target=/dist,ro \
-    uv pip install /dist/${WHEEL}[server]
+RUN adduser --disabled-password --gecos "" aipscan
```
replaceafill, line 45: What's the advantage of having a specific user in this new approach?

